PR #9, opened 2024-08-19
Update to Lean v4.10.0

MackieLoeffel:
This MR updates the code to lean v4.10.0. This includes changing Std to Batteries and adapting the proofs. This is the first time I am working with Lean, so I don't know where the errors were coming from and if I fixed them in the right way. In particular, I am confused why the change to [src/Iris/BI/Notation.lean](https://github.com/leanprover-community/iris-lean/compare/master...MackieLoeffel:iris-lean:port_to_v4.10.0?expand=1#diff-5e709d9df7576570f685bcb0d8fe32489bbc10db582eefdc38f50c3824a4506c) was necessary.


lakefile.lean
```diff
@@ -1,11 +1,11 @@
 import Lake
 open Lake DSL
 
-package iris where
+package Iris where
```
digama0, line 4:
```suggestion
package iris where
```
Convention for package names uses lowercase, although you are right that lean_libs should be pascal case.

src/Iris/BI/Notation.lean
```diff
@@ -31,7 +31,13 @@ partial def unpackIprop [Monad m] [MonadRef m] [MonadQuotation m] : Term → m T
...
   | `(($P))                  => do `(($(← unpackIprop P)))
   | `($P $[ $Q]*)            => do ``($P $[ $Q]*)
-  | `(if $c then $t else $e) => do `(if $c then $(← unpackIprop t) else $(← unpackIprop e))
+  | `(if $c then $t else $e) => do
+    -- TODO: Why does `(if $c then $(← unpackIprop t) else $e) give
+    -- "invalid use of `(<- ...)`, must be nested inside a 'do'
+    -- expression" here, but not in the other cases?
+    let t ← unpackIprop t;
+    let e ← unpackIprop e;
+    `(if $c then $t else $e)
```
digama0, line 40:
```suggestion
  | `(if $c then $t else $e) => do
    let t ← unpackIprop t
    let e ← unpackIprop e
    `(if $c then $t else $e)
```
This is a known issue, it triggers because we are using `<-` inside an `if` statement body, which is a footgun in normal usage. It's a false positive here because the `if` is actually in a quotation so it isn't actually running, but the check which produces the error misses this issue. We hit several instances of this when the error was first rolled out in user code, and decided that it's best just to work around it for the present and do as you did here.

src/Iris/ProofMode/Instances.lean
```diff
@@ -511,7 +511,7 @@ instance fromPure_pure_sep_true (a1 a2 : Bool) (φ1 φ2 : Prop) [BI PROP] (P1 P2
     | true, false => persistent_and_affinely_sep_l.1
     | true, true => affinely_and.1.trans persistent_and_sep_1
 
-instance fromPure_pure_wand_true (a : Bool) (φ1 φ2 : Prop) [BI PROP] (P1 P2 : PROP)
+instance fromPure_pure_wand_true (_ : Bool) (φ1 φ2 : Prop) [BI PROP] (P1 P2 : PROP)
```
digama0, line 514:
```suggestion
instance fromPure_pure_wand_true (φ1 φ2 : Prop) [BI PROP] (P1 P2 : PROP)
```
This looks like a mistake.

src/Iris/ProofMode/Tactics/Clear.lean
```diff
@@ -20,7 +20,7 @@ def clearCore {prop : Q(Type)} (_bi : Q(BI $prop)) (e e' out goal : Q($prop))
   if out.isAppOfArity ``intuitionistically 3 then
     have out' : Q($prop) := out.appArg!
     have : $out =Q iprop(□ $out') := ⟨⟩
-    pure q(clear_intuitionistic $pf)
+    pure q(clear_intuitionistic (Q:=$goal) $pf)
```
digama0, line 23:
```suggestion
    pure q(clear_intuitionistic (Q := $goal) $pf)
```

lakefile.lean
MackieLoeffel: Thanks for the explanation. I updated the MR.

src/Iris/BI/Notation.lean
MackieLoeffel: I see, thanks! I've removed the comment.

PR #11, opened 2025-02-13
feat: Alternative definition for CMRA

markusdemedeiros:
Making this PR to request a comment (cc @digama0). 

A recent [paper](https://www.danielgratzer.com/papers/idempotent-resources-in-separation-logic.pdf) by Gratzner, Møller and Birkedal proposes a simplified definition for CMRAs, which removes the need for the ``core`` operator. They have [fully ported Iris](https://gitlab.mpi-sws.org/adamAndMath/iris/-/tree/fossacs-2025?ref_type=tags) with these changes, so it would be realistic for iris-lean to use their definition as well. See in particular section 6.2 of the linked paper. 

Which definition of ``CMRA`` should the project use moving forward?

### DIFF
--- a/src/Iris/Algebra/CMRA.lean
+++ b/src/Iris/Algebra/CMRA.lean
@@ -8,46 +8,63 @@ import Iris.Algebra.OFE
 namespace Iris
 open OFE
 
-class CMRA (α : Type _) extends OFE α where
-  pcore : α → Option α
+class PreCMRA (α : Type _) extends OFE α where
   op : α → α → α
   validN : Nat → α → Prop
   valid : α → Prop
 
-  op_ne : NonExpansive (op x)
-  pcore_ne : x ≡{n}≡ y → pcore x = some cx →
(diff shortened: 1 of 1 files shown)


digama0:
I already saw that this change is coming, but my intention is to wait until it is merged upstream before changing iris-lean. (There may be other practical reasons that the patch cannot be merged or for it to be delayed.) But feel free to make a patch so that it is easy to upgrade on our end, if you want.

markusdemedeiros:
Makes sense, thanks!

srghma:
seems like they wont implement this ([older](https://gitlab.mpi-sws.org/adamAndMath/iris/-/compare/master...fossacs-2025?from_project_id=118), [newer](https://gitlab.mpi-sws.org/adamAndMath/iris/-/compare/master...fossacs?from_project_id=118)) change (IIUC because it would **break or disallow** valid and useful Resource Algebras (RAs) like $M_{\text{disintegrable}}$)

https://gitlab.mpi-sws.org/iris/iris/-/work_items/224#note_127483

PR #13, opened 2025-03-12
feat: More results on OFEs

quangvdao:
I added some more lemmas and definitions for OFEs, as a first pass to familiarize myself with the notion.

Suggestion: should we rename some definitions to follow mathlib convention? For instance, function fields should be `toFun`.


digama0:
can you move the bump to a separate PR?

digama0:
also this project runs on the stable, not the RC, unless you have a good reason?

quangvdao:
Got it, will revert

src/Iris/Algebra/OFE.lean
```diff
@@ -241,6 +379,7 @@ instance : COFE Unit where
   compl _ := ()
   conv_compl := ⟨⟩
 
+/-- Functors from COFEs to OFEs -/
```
digama0, line 382: These aren't actually functors from COFEs to OFEs, they are profunctors, aka bifunctors which are contravariant on the left, aka functors `COFEᵒᵖ × COFE ⥤ OFE`.

digama0 (approved):

PR #15, opened 2025-03-16
feat: Define `HeapLang`

quangvdao:
This PR defines `HeapLang`, as defined in the `iris_heap_lang/lang.v` file in the Rocq development. We use `HashMap` for the heap (but this could be changed).

We try to stay faithful to the original code, up to renaming definitions to align with Lean convention.


CatsAreFluffy:
Logical operators applied to integers should do bitwise operations ([source](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris_heap_lang/lang.v#L525)).

PR #16, opened 2025-03-17
Update toolchain to v4.17.0

Shreyas4991:


digama0:
you should remove the `lakefile.lean`

Shreyas4991:
> you should remove the `lakefile.lean`

Done

PR #20, opened 2025-03-30
UPred 

markusdemedeiros:
Realistically speaking I think the CMRA lemmas need to be finished before more progress can happen here

### DIFF
--- a/src/Iris/Algebra/uPred.lean
+++ b/src/Iris/Algebra/uPred.lean
@@ -1,0 +1,73 @@
+/-
+Copyright (c) 2025 Leo Stefanesco. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Markus de Medeiros
+-/
+
+import Iris.Algebra.CMRA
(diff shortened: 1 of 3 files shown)


markusdemedeiros:
This can start being reviewed, if you want, and I'm happy to spend more time gardening the code if you have advice for that. There are currently two sorry's left which I've asked about [here](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/.60pcore_mono.60.20vs.20.60pcore_mono'.60/near/511842114).

This PR includes the following features
- All of @suhr 's CMRA lemmas  
- `RFunctor`, `RFunctorContractive`, `URFunctor`, `RFunctorContractive` 
- UCMRA and functor for discrete functions 
- UCMRA and functor for `OptionO`
- CMRA and functor `gen_map`, which has the same `OFE` structure as `gmap` (though it does not re-create the `gmap` datatype)
- Definition of `uPred`, plus `BI` and `BILaterContractive` instance 
- Definition of `iProp`

PR #24, opened 2025-04-14
CMRA lemmas for iProp 

markusdemedeiros:


src/Iris/Algebra/OFE.lean
```diff
@@ -138,6 +142,10 @@ class Discrete (α : Type _) [OFE α] : Prop where
 theorem Discrete.discrete_n [OFE α] [Discrete α] {n} {x y : α} (h : x ≡{n}≡ y) : x ≡ y :=
   Discrete.discrete_0 (OFE.Dist.le h (Nat.zero_le _))
 
+class Leibniz (α : Type _) [OFE α] : Prop where
```
digama0, line 145: I don't like this name, I think we should rename it to `Tight` or similar. It's called this in Rocq because they call `=` "leibniz equality" to differentiate it from other kinds of equality like setoid equality or definitional equality, but this terminology is not used in lean so I think it will just be confusing.

PR #40, opened 2025-05-15
CI for build

Shreyas4991:
Adds CI to build the project for each commit on main and each PR.


.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
...
+
+# Cancel previous runs if a new commit is pushed to the same PR or branch
+concurrency:
+  group: ${{ github.ref }}  # Group runs by the ref (branch or PR)
+  cancel-in-progress: true  # Cancel any ongoing runs in the same group
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages and modify PR labels
+permissions:
+  contents: read          # Read access to repository contents
+  pages: write            # Write access to GitHub Pages
```
digama0, line 31: :raised_eyebrow:

digama0:
This seems overcomplicated, surely we can just call lean-action here?

Shreyas4991:
> This seems overcomplicated, surely we can just call lean-action here?

It doesn't do the label management we have found useful so far. Also, I recall that lean-action was largely for mathlib-downstream projects

digama0:
I don't want to be using a baroque custom setup. Please use lean-action here.

Shreyas4991:
@digama0  : This PR is ready to be checked

.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
+name: Build Iris Lean
+
+on:
+  push:
+    branches: ["master"] # replace "main" with the default branch
```
digama0, line 5:
```suggestion
    branches: ["master"]
```

PR #42, opened 2025-05-24
Port proofmode theory

markusdemedeiros:


src/Iris/ProofMode/Modalities.lean
```diff
@@ -0,0 +1,55 @@
...
+Copyright (c) 2025 Markus de Medeiros. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Markus de Medeiros
+-/
+import Iris.BI
+
+namespace Iris.ProofMode
+open Iris.BI
+
+inductive ModalityAction (PROP1 PROP2 : Type u) : Type u where
```
digama0, line 11: @markusdemedeiros I put in the universes here because it may not be obvious that they are required to be the same by this definition. We may need to revisit this; I do not really understand the purpose of this definition but I suspect this is some coq universe shenanigans we'll have to do differently. Let's discuss this on zulip.

PR #47, opened 2025-06-12
feat: UPred Soundness 

markusdemedeiros:
Basic UPred soundness lemmas, plus some nearby one-liners that were easy to port. 

The other UPred soundness lemmas (w/ the various kinds of updates, internal eq, etc) reduce to these cases here, so this should be merged before those can be finished.


digama0:
@markusdemedeiros This PR had some conflicts, some of the same theorems got added in another PR by different names. Please double check that the names and theorem statements ended up as intended.

Issue #27, opened 2025-05-14
Local Updates


markusdemedeiros:
- Definitions and lemmas for local updates. 
- Instances of local updates for basic CMRA's (unit, product, discrete functions, and option). 

All dependencies of [local_updates.v](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/local_updates.v?ref_type=heads) have been ported.

PR #49, opened 2025-06-12
Add Local Updates

suhr:
See #27. Uses #66.

Do we have CMRAs for `Unit` and products?

### DIFF
--- a/src/Iris/Algebra/CMRA.lean
+++ b/src/Iris/Algebra/CMRA.lean
@@ -169,8 +169,19 @@ theorem _root_.Iris.OFE.Equiv.opM {x₁ x₂ : α} {y₁ y₂ : Option α}
     (H1 : x₁ ≡ x₂) (H2 : y₁ ≡ y₂) : x₁ •? y₁ ≡ x₂ •? y₂ :=
   equiv_dist.2 fun _ => H1.dist.opM H2.dist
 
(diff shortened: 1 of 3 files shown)


markusdemedeiros:
Good catch, I don't actually think we do. Feel free to add them (`Unit` should be like 1 line using `ofDiscrete`)

suhr:
In `local_updates.v`, `alloc_option_local_update` is proved like so:

```rocq
  Lemma alloc_option_local_update {A : cmra} (x : A) y :
    ✓ x →
    (None, y) ~l~> (Some x, Some x).
  Proof.
    move=>Hx. apply local_update_unital=> n z _ /= Heq. split.
    { rewrite Some_validN. apply cmra_valid_validN. done. }
    destruct z as [z|]; last done. destruct y; inversion Heq.
  Qed.
```

But in Lean, the proof turned out to be vacuous:

```lean
  theorem alloc_option_local_update {α : Type} [CMRA α]
      {x : α} (y : α) (_: ✓ x): (none, y) ~l~> (some x, some x) :=
    fun _ mz _ e =>
      match mz with
      | .none           => False.elim e
      | .some .none     => False.elim e
      | .some (.some _) => False.elim e
```

That's sus, is the instance of `CMRA (Option α)` defined correctly?

suhr:
Fixed this, and now the PR is ready for review.

markusdemedeiros:
Nice work! Mario is too busy to review and merge PR's right now but if you wanted to do more, #58 should be very accessible to you.

PR #56, opened 2025-06-22
feat: Frac

markusdemedeiros:
cc @Shreyas4991


Shreyas4991:
Shall I make some commits. I can see one or two things that could be improved.

markusdemedeiros:
Go for it! You can own this PR now. I just needed the definitions to make forward progress.

markusdemedeiros:
Depends on #72

Issue #37, opened 2025-05-14
Extensible vectors of RFunctors


markusdemedeiros:
In Iris, one typically specifies that a particular RFunctor is contained in a global list of locally contractive RFunctors Σ (so-called "GFunctors"). Iris accomplishes this using the `inG`  (from [own.v](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/lib/own.v)) and `subG` (from [iprop.v](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/lib/iprop.v)) typeclasses.

This is not yet implemented in iris-lean: we currently require that the vector of RFunctors be supplied explicitly, and that there is a typeclass instance specifying that all of its elements are locally contractive. 

Porting the `gFunctors` typeclasses from Rocq to use unbundled types may work. This would entail porting `subG` and `inG` as above, and providing the right typeclass instances so that Lean knows every RFunctor in the list is locally contractive. We could also consider writing a metaprogram to manage the `GFunctors`, since they are not a part of the logic. 

This PR depends on PR #25 which defines iProp.

PR #76, opened 2025-07-14
feat: Vectors of rFunctors

Remyjck:
Here is an ongoing attempt to add extensible vectors of rFunctors #37


markusdemedeiros:
Great start on this :) 

One thing I'm thinking about: I'm replacing `GenMapOF` in my heaps PR to support heaps with infinitely many allocated elements. I'm doing this for other reasons, but because the iProp construction is based on  `GenMapOF`, I'm wondering how much the rest of the `inG` machinery depends on the number of pieces of ghost state being finite?

Remyjck:
I think/hope that as long as you carry around a witness that your index is valid (what is being done by `GId` here), then it should be fine to have infinite pieces of ghost state.

PR #107, opened 2025-12-16
Unify handling of proof mode terms

MackieLoeffel:
This PR unifies the handling of proof mode terms following the Rocq implementation: It first introduces a tactic `ispecialize` to specialize a given Iris hypothesis with a proof mode term. Then it introduces an `ihave` tactic that allows to introduce a pure entailment into the proof mode. Tactics that use a proof mode term can then simply call the `iHave` function provided by `ihave` to parse a given proof mode term. This is then used to refactor `iapply` to a sequence of `ihave` and `ispecialize` following the Rocq implementation.

There are open TODOs about changing `iapply` to be able to introduce and instantiate mvars (see the tests). This is currently blocked on the use of type class search, see [#iris-lean > Issues with typeclasses in the proof mode](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Issues.20with.20typeclasses.20in.20the.20proof.20mode/with/563410548).

Additionally, this PR does some small refactorings:
- It makes ⊢ a definition `EmpValid` like in Rocq instead of just a notation. This prevents typeclass problems for `IntoEmpValid`.
- It introduces the `⊢@{PROP}` like in Rocq to specify the bi of an entailment.
- It adds the `Goals` type for collecting the goals generated by a tactic and automatically putting dependent goals last.

### DIFF
--- a/src/Iris/BI/Lib/BUpdPlain.lean
+++ b/src/Iris/BI/Lib/BUpdPlain.lean
@@ -48,26 +48,22 @@ theorem BUpdPlain_mono {P Q : PROP} : (P ⊢ Q) → (BUpdPlain P ⊢ BUpdPlain Q
     iapply H
     apply H
   iintro ⟨Ha, H2⟩
-  ispecialize Ha HQR
-  iapply Ha
+  iapply Ha $! HQR
   iapply H1
   iexact H2
 
 theorem BUpdPlain_idemp {P : PROP} : BUpdPlain (BUpdPlain P) ⊢ BUpdPlain P := by
   unfold BUpdPlain
   iintro Hp R H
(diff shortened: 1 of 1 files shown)


lzy0505:
The failing test `exact_forall'` actually worked before (with an extra `istart`), and 
```
theorem apply_forall4 [BI PROP] (P Q : α → PROP) (a : α) : (∀ x, P x -∗ Q x) ⊢ P a -∗ Q a := by
  iintro H HP
  iapply H $! ?_ with HP
```
worked before, but now fails.

Seems like we did a slightly better job at instantiating mvars. But I don't know if we should make some ad-hoc fixes for these cases, or wait until we have a better general solution.

PR #117, opened 2026-01-15
iProp cleanup

Remyjck:
Went through the instance proofs for iProp #97 and refactored them into intermediate lemmas.

N.B. : I tried to reprove the `own_alloc_strong_dep` theorem from the rocq formalization, but it seems like this relies on gmaps having finite support.


markusdemedeiros:
Could you please PR this into the main branch instead?

Remyjck:
Should be done :slightly_smiling_face:

PR #114, opened 2026-01-12
Add ProofModeM monad

MackieLoeffel:
Builds on #109 

Add a ProofModeM monad that automates the goal management of proof
mode tactics. This subsumes the previous `Goals` structure.

PR #119, opened 2026-01-20
refactor: Clean up tactic implementations 

MackieLoeffel:
Builds on #114 

- Collect everything related to Hyps in Expr.lean
- Make all proof mode theorems private
- Deduplicate code in icases
- Add error messages for failing typeclass searches
- Tweak icases such that the code path does not depend on the result of
TC synthesis
- Define iintuitionistic and ispatial as special cases of icases
- Add test cases for failures
- Introduce a type for intropatterns
- Force forall introduction to use a pure intro pattern, similar to
cases
- Add more precise error location for intro patterns
- Add proof mode term support for icases
- Significantly simplify iassumption


lzy0505:
Please let me know when the PR is ready. I will review it.

src/Iris/ProofMode/Tactics/Pure.lean
```diff
@@ -86,16 +77,18 @@ elab "ipure_intro" : tactic => do
...
-    let _ ← synthInstanceQ q(Affine $e)
+    let .some _ ← trySynthInstanceQ q(Affine $e)
+      | throwError "ipure_intro: context is not affine"
     mvar.assign q(pure_intro_affine (P := $e) (Q := $goal) $m)
   | .const ``false _ =>
     have : $b =Q false := ⟨⟩
     mvar.assign q(pure_intro_spatial (P := $e) (Q := $goal) $m)
-  | _ => throwError "failed to prove `FromPure _ {goal} _`"
-  addMVarGoal m.mvarId!
+  | _ => throwError "ipure_intro: {b} must be either true or false"
```
lzy0505: Is this wildcard case necessary?

src/Iris/ProofMode/Tactics/Cases.lean
```diff
@@ -275,15 +221,20 @@ partial def iCasesCore
...
 
   -- process pattern
   let pf2 ← iCasesCore bi hyps' goal b A A' h pat (λ hyps => addBIGoal hyps goal)
 
-  mvar.assign q(($pf).1.trans $pf2)
+  mvar.assign q(($pf).trans (($pf').1.trans $pf2))
+
+-- TODO: are these shortcuts necessary?
+macro "iintuitionistic" hyp:ident : tactic => `(tactic | icases $hyp:ident with □$hyp:ident)
+macro "ispatial" hyp:ident : tactic => `(tactic | icases $hyp:ident with ∗$hyp:ident)
```
lzy0505: I would never use ispatial.. iintuitionistic is handy but too long to type.

MackieLoeffel: Let's see whether they are useful and if not, we can remove them at some point.

src/Iris/ProofMode/Tactics/Pure.lean
MackieLoeffel: Yes, `b` might be not syntactically true or false if the type class instances that generate it do something weird. I added a comment that this indicates a bug in the instances.

src/Iris/Tests/Tactics.lean
```diff
@@ -217,14 +285,11 @@ example [BI PROP] (Q : PROP) : □ Q ⊢ Q := by
...
-example [BI PROP] (Q : PROP) : <affine> ⌜⊢ Q⌝ ⊢ Q := by
-  iintro ⌜H⌝
+/- Tests `iassumption` failure -/
+/-- error: iassumption: no matching assumption -/
+#guard_msgs in
+example [BI PROP] (P Q : PROP) : □ P ⊢ Q := by
+  iintro □_HQ
   iassumption
 
 end assumption
```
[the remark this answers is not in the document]
MackieLoeffel: Indeed, good suggestion. I added the test you suggested.

Issue #116, opened 2026-01-15
Asserting iProps with `ihave`


lzy0505:
We often write `iAssert (P : PROP) with "[R1 R2]" as "HP”` in Iris-Rocq.
It is currently not possible to do this with `ihave`.

One can do 
```
have H : ((R1 ∗ R2) ⊢ P) := by ...
ihave H $$ [R1 R2]
```
to achieve something similar.

PR #120, opened 2026-01-20
feat: Add ihave : Q tactic 

MackieLoeffel:
Builds on #119 

Add the ihave : Q tactic that allows asserting new separation logic
hypothesis

Closes https://github.com/leanprover-community/iris-lean/issues/116


markusdemedeiros:
I think we're all good to go I'll just do it

PR #160, opened 2026-02-20
Finite Maps 

markusdemedeiros:
## Description
More code derived from #131. This version generalizes many of the constructions to work with PartialMaps (where appropriate), eliminates several typeclasses as they can be derived as theorems. A `TreeMap` instance is provided. 

The `kmap` and `mapSeq` ports are not yet complete but I would like to get feedback on this. There are also two theorems with universe issues we should solve at some point.

Fixes # (issue number)

## Checklist

## Generative AI Guidelines

### DIFF
--- a/src/Iris/Std/HeapInstances.lean
+++ b/src/Iris/Std/HeapInstances.lean
@@ -290,13 +290,13 @@ private theorem get?_foldl_alter_impl_sigma {l : List ((_ : K) × V)}
         ((l.find? (fun x => (compare x.1 k).isEq)).map (fun kv => (kv.1, kv.2))) := by
   induction l generalizing init with
   | nil =>
-    simp [foldl_nil]
+    simp [List.foldl_nil]
   | cons hd tl IH =>
-    rw [foldl_cons, IH (WF.constAlter! hinit) (hl.tail), Const.get?_alter! hinit]
+    rw [List.foldl_cons, IH (WF.constAlter! hinit) (hl.tail), Const.get?_alter! hinit]
     by_cases h : compare hd.1 k = .eq <;> simp [h]
     rw [← Const.get?_congr hinit h]
     have hhead_none : tl.find? (fun x => (compare x.1 k).isEq) = none := by
-      refine find?_eq_none.mpr fun _ hkv He => rel_of_pairwise_cons hl hkv ?_
+      refine List.find?_eq_none.mpr fun _ hkv He => List.rel_of_pairwise_cons hl hkv ?_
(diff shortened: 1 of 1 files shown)


markusdemedeiros:
Thanks for the feedback! 

> Stepping back, what will this interface be used for? Do we truly need generality? For example, I can’t think of any scenarios where users need a custom map for big operations. Why can't we go more extreme, just using a concrete implementation?

Ya I grepped around Iris and I don't see anywhere in Iris that uses truly finite maps outside of defining the big ops, and the gset CMRA. Now unless I missed something, all of the lemmas from your big ops PR (except those regarding kmap and mapSeq atm) should be present here in some form, so since your big ops PR already builds off of this interface I'd imagine the theory is rich enough as is? In my experience I mostly interact with the BigOps API instead of the underlying map, so any definition for map/filter/etc will be interchangeable, but perhaps your experience differs here. 

Big ops do fix a concrete type (list) and I do think that's the right way to go, there's no reason to define a separate BigOp for every FiniteMap when a FiniteMap defines a list. It seems really common for mathlib and stdlib types to implement a toList function, and that's exactly what my interface requires to turn a partial map into a finite map, so I like how the two work together. Like I said at some point before I just want to make sure it's easy to take your favorite finite type and do a big conjunction over it, and I think this structure accomplishes that.

> However, I’m concerned that we’ll need to prove lots of lemmas for all the operations we define ourselves to make the interface usable, if the users are expected to interact with it most of the time. It seems to be a redundant work, given that the interface instances might already have their own implementations and associated lemmas that we can just use. 

Point taken, but by my reading the lemmas about big ops are relatively simple and I think this interface is rich enough to cover them. Do you have a different view? Like, do you find yourself relying on properties that only hold on `gmap` often or do you (like me) just use the basic BigOp lemmas? 

> This is why I suggested before that we should consider some commonly used implementations when designing the interface. For example, we could fix an implementation like TreeMap and try to reuse its existing theories as much as possible, only redefining operations when necessary.

The problem with this is that if you don't redefine the operations, either need to fix a single type, or you need to specify all of the equations, and I think we agree that that is something we should avoid. 

I don't like the idea of saying that TreeMaps are the only finite map. Now, if there is a need for some big-op-over-TreeMap specific lemmas I think that could be a fine idea, just like gmap does for Iris-Rocq. From my perspective, I want to like, take the separating conjunction over a finitely-supported probability distribution and not worry about if that type has a good analogue of `zip`. This design lets you do that.

Sorry for the essay. What are your thoughts? Do you have any more concrete issues or suggestions?

Issue #69, opened 2025-07-01
iRevert tactic


oliversoeser:
Port the `iRevert` tactic

PR #175, opened 2026-03-17
feat: refactor irevert implementation

HaleOIC:
## Description
* Fixes # [69](https://github.com/leanprover-community/iris-lean/issues/69)
* Add `Make` related type class and instances 
* Add relative complete tactic for `irevert`

### DIFF
--- a/PORTING.md
+++ b/PORTING.md
@@ -271,18 +271,18 @@ Some porting tasks will require other tasks as dependencies, the GitHub issues p
   - [ ] AddModal
   - [x] IntoLater
 - [ ] `class_instances_make.v`
-  - [ ] Affine 
-  - [ ] Absorbing
-  - [ ] Embed
-  - [ ] Sep
-  - [ ] And 
-  - [ ] Or 
-  - [x] Affinely
-  - [x] Absorbingly
-  - [x] Persistenly
-  - [x] Intuitionistically
-  - [x] Later
-  - [ ] Except-0
+  - [ ] QuickAffine
+  - [ ] QuickAbsorbing
+  - [ ] MakeEmbed
+  - [ ] MakeSep
+  - [ ] MakeAnd
+  - [ ] MakeOr
+  - [x] MakeAffinely
+  - [x] MakeIntuitionistically
+  - [x] MakeAbsorbingly
+  - [x] MakePersistenly
+  - [x] MakeLaterN
+  - [ ] MakeExcept0
 - [ ] `class_instances_plainly.v` (InstancesPlainly.lean)
   - [x] basic instances
   - [x] FromModal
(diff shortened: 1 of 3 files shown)


src/Iris/ProofMode/Tactics/Revert.lean
```diff
@@ -0,0 +1,136 @@
...
+      | throwError "irevert: variable not in scope"
+
+    let v : Level ← Meta.getLevel ldecl.type
+    have α : Q(Sort v) := ldecl.type
+
+    let (_, mvarId) ← mvar.revert #[f]
+    mvarId.withContext do
+      if ← Meta.isProp α then
+        have φ : Q(Prop) := α
+        let p : Q($prop) := q(iprop(<affine> ⌜$φ⌝))
```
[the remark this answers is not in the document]
HaleOIC: You’re right. This part was too strict, and it made the refactor get stuck for a while. Please take a look at the latest version.

src/Iris/ProofMode/Tactics/Revert.lean
```diff
@@ -0,0 +1,169 @@
...
+    if ← Meta.isProp α then
+      have φ : Q(Prop) := α
+      st.revertLeanPropHyp f φ
+    else
+      st.revertLeanForallHyp f α
+
+elab "irevert" hs:(colGt ident)+ : tactic => do
+  ProofModeM.runTactic fun mvar { bi, e, hyps, goal, .. } => do
+    let init : RevertState bi e goal :=
+      { e := e, hyps := hyps, goal := goal, reverted := #[], pf := q(id) }
```
MackieLoeffel: You should be able to write this as `{ e, hyps, goal, pf := q(id) }`

src/Iris/ProofMode/Tactics/Revert.lean
```diff
@@ -53,84 +69,101 @@ private theorem pure_revert [BI PROP] {Δ P Q : PROP} {φ : Prop}
...
+  let x : Q($α) := mkFVar f
+  let ΦExpr ← mkLambdaFVars #[x] goal
+  let some Φ ← checkTypeQ ΦExpr q($α → $prop)
+    | throwError "irevert: failed to construct forall body"
+  let goal' : Q($prop) := q(BI.forall $Φ)
+  let pf'Expr ← withLocalDeclDQ `h q($e ⊢ BI.forall $Φ) fun h => do
+    let step ← mkAppM ``forall_revert #[h]
+    let oldGoalProof ← mkExpectedTypeHint (mkApp step x) q($e ⊢ $goal)
+    mkLambdaFVars #[h] (mkApp pf0 oldGoalProof)
+  let some pf' ← checkTypeQ pf'Expr q(($e ⊢ $goal') → ($origE ⊢ $origGoal))
```
MackieLoeffel: Typechecking is an expensive operation and should be avoided when it is not necessary. Is every a case where this can fail? If not, you can use `have` instead. (same comment for `checkTypeQ ΦExpr q($α → $prop)` above)

PR #307, opened 2026-04-09
feat: rocq_alias accepts the declaration name as alias

ayhon:
## Description
Makes `rocq_alias` accept `↓` instead of an identifier. In this case, the declaration name is used as an identifier.


ayhon:
As discussed in [Zulip](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Noting.20Rocq.20names.20when.20porting/near/584416445), there is a concern with this syntax becoming outdated if the name changes. A possible fix is only using the `↓` to ask `rocq_alias` to suggest the declaration name as the Rocq alias.

ayhon:
Actually, `rocq_alias ↓` is doubly dangerous, since the `declName` includes the namespaces!

PR #309, opened 2026-04-10
feat: add plainly definitions

### LINKED ISSUE #103: Complete porting Plainly
https://github.com/leanprover-community/iris-lean/blob/master/src/Iris/BI/Plainly.lean is not fully done, which caused some issues in https://github.com/leanprover-community/iris-lean/pull/102

It will be nice to check if we ported the most relevant lemmas and instances of the derived connectives

### LINKED ISSUE #266: Port bi/plainly.v
## Summary

Port the remaining items from [`bi/plainly.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/bi/plainly.v).

## Subtasks

- [ ] Big op lemmas
- [ ] Internal eq lemmas
- [ ] Other lemmas (e.g. about except0 and laterN)

## Already ported

- [x] Plainly lemmas

## Dependencies

**Rocq dependencies:**

- #260
- #222
- #250

ayhon:
## Description
Adds various definitions to `Plainly.lean`. Also puts the `[BI PROP]` condition in `DerivedLawsLater`'s lemmas into a `variable` statement (which is the main reason there are so many changes in that file).

Addresses #103, although some definitions have not been ported yet. 

Other definitions, I decided were not in scope. I have collected these definitions with my reasoning behind my decision in the following.

## Theorems involving `bigSepS` and `bigSepM2`
These operations haven't been implemented in Iris-Lean yet. I believe it makes sense adding them after some other PR.

### Instances of `Proper`.
If I understood correctly, these instances are mainly used to allow Rocq's generalized rewriting system to work. Since we don't have any equivalent in Lean, I have left them out.

For example:
```
  Global Instance plainly_if_proper p : Proper ((⊣⊢) ==> (⊣⊢)) (@plainly_if PROP _ p).
  Proof. solve_proper. Qed.
  Global Instance plainly_if_mono' p : Proper ((⊢) ==> (⊢)) (@plainly_if PROP _ p).
  Proof. solve_proper. Qed.
  Global Instance plainly_if_flip_mono' p :
    Proper (flip (⊢) ==> flip (⊢)) (@plainly_if PROP _ p).
  Proof. solve_proper. Qed.
```

## `WeakMonoidHomomorphism` of `flip`s
We don't really have a `flip` function in Iris Lean right now, and all other definitions which mention it perform the flip "inline", not through a function. I therefore questioned their usefulness, and left them out, although I'll be happy to implement them if we think they should be included.

For example:
```
  Global Instance plainly_sep_entails_weak_homomorphism :
    WeakMonoidHomomorphism bi_sep bi_sep (flip (⊢)) (@plainly PROP _).
  Proof. split; try apply _. intros P Q; by rewrite plainly_sep_2. Qed.
  Global Instance plainly_sep_entails_homomorphism `{!BiAffine PROP} :
    MonoidHomomorphism bi_sep bi_sep (flip (⊢)) (@plainly PROP _).
  Proof. split; try apply _. simpl. rewrite plainly_emp. done. Qed.
```

## `prop_ext_2`
It seems like the `prop_ext_2` rocq alias was being used in [`Iris.Bi.SiProp`](https://github.com/leanprover-community/iris-lean/blob/master/src/Iris/BI/SIProp.lean#L245-L247). I'm not sure if this alias is correct or not, but since `prop_ext_2` is equivalent to `prop_ext.2`, I decided it would be fine to do without it.

____

Some other interesting addition is `BiEntails.ofMono`, which proves the monotonicity of some modality for `BiEntails`, given a proof for `Entails`. I found I needed it in enough places that a new definition was warranted.

## Checklist

### DIFF
--- a/src/Iris/Std/Positives.lean
+++ b/src/Iris/Std/Positives.lean
@@ -349,6 +349,17 @@ class Countable (A : Type) where
 instance some_inj {A} : Iris.Std.Injective (@some A) where
   inj _ _ := by rintro ⟨⟩; rfl
 
+instance {A} [Countable A] : DecidableEq A := by
+  open Countable in
+  intros x y
+  if h : encode x = encode y then
+    apply Decidable.isTrue
+    apply some_inj.inj
+    rw [←decode_encode, ←decode_encode, h]
+  else
+    apply isFalse
+    grind only
+
 instance encode_inj [c : Countable A] : Iris.Std.Injective (c.encode) where
   inj x _ Hxy := by
     apply some_inj.inj
--- a/src/Iris/Std/Positives.lean
+++ b/src/Iris/Std/Positives.lean
@@ -349,17 +349,6 @@ class Countable (A : Type) where
 instance some_inj {A} : Iris.Std.Injective (@some A) where
   inj _ _ := by rintro ⟨⟩; rfl
 
-instance {A} [Countable A] : DecidableEq A := by
-  open Countable in
-  intros x y
-  if h : encode x = encode y then
-    apply Decidable.isTrue
-    apply some_inj.inj
-    rw [←decode_encode, ←decode_encode, h]
-  else
-    apply isFalse
-    grind only
-
 instance encode_inj [c : Countable A] : Iris.Std.Injective (c.encode) where
   inj x _ Hxy := by
     apply some_inj.inj
--- a/PORTING.md
+++ b/PORTING.md
@@ -234,7 +234,7 @@ Some porting tasks will require other tasks as dependencies, the GitHub issues p
 - [ ] `plainly.v`
   - [x] plainly lemmas
   - [ ] big op lemmas
-  - [ ] internal eq lemmas
+  - [x] internal eq lemmas
 - [ ] `sbi.v`
 - [ ] `sbi_unfold.v`
 - [ ] `telescopes.v`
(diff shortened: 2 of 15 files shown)


src/Iris/Std/Positives.lean
```diff
@@ -349,6 +349,17 @@ class Countable (A : Type) where
 instance some_inj {A} : Iris.Std.Injective (@some A) where
   inj _ _ := by rintro ⟨⟩; rfl
 
+instance {A} [Countable A] : DecidableEq A := by
```
markusdemedeiros: I think this is not a good instance to add, since it will fire for other countable types that already have DecidableEq instances. Is there a reason you want this?

markusdemedeiros:
Overall: a super nice PR! There are lots of little changes I made, I'd suggest you read the git diffs and ask me if you have any questions. The biggest thing I'd really suggest is that you try to `apply?` before writing a new proof: four of your big op lemmas were just instances of more general big op lemmas.

src/Iris/Std/Positives.lean
ayhon: I suppose you closed this since you figured the solution already, but I wanted to have Decidable Eq K for maps while just assuming Countable K. I guess we could add a (prio := low) or similar to the instance. Wdyt?

ayhon: Ok, nvm, I saw the fix.

PR #328, opened 2026-04-21
feat: Multi-package repository 

### LINKED ISSUE #122: Add Iris-Lean-Math as a project in this repo
This is blocked by manual registration to Reservoir. See the [zulip thread]([#general > Multiple packages in the same repo + Resevoir?](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Multiple.20packages.20in.20the.20same.20repo.20.2B.20Resevoir.3F/with/569059510)).

markusdemedeiros:
## Description
Moves Iris-Lean to a multi-package repo with the core code living in IrisLean, and math code living in IrisLeanMath. 

This PR is a WIP so I can test the CI, and test that the packages can be correctly imported. 

Fixes #122


IrisLeanMath/lake-manifest.json
lzy0505, line 1: Are these all necessary? Should this file sync with the toml file?

IrisLeanMath/.gitignore
```diff
@@ -0,0 +1 @@
+/.lake
```
lzy0505, line 1: Seems redundant.

IrisLeanMath/lake-manifest.json
markusdemedeiros: I think most of them come from the Iris dependency, no?

markusdemedeiros: (It is auto-generated from the toml file if that's what you're asking)

IrisLeanMath/.github/workflows/create-release.yml
lzy0505, line 1: Not sure if you can have GitHub CI files in non-root directories.

markusdemedeiros: Good catch. Auto-generated by lake and I missed it.

IrisLeanMath/lean-toolchain
```diff
@@ -0,0 +1 @@
+leanprover/lean4:v4.29.0
```
lzy0505, line 1: Questions: 
1. Do the two libraries always use the same Lean version? 
2. I remember you mentioned that you wish to use the mathlib nightly build. How can this work?

markusdemedeiros: 1. Yes, the plan would be to bump them at the same time. 
2. The idea is that if a project _already_ uses nightly, they can import Iris-Lean without conflicting against the Iris-Lean-Math dependency. But the core math library we maintain should use the stable version so we only have to update it every 6 weeks or so.

scripts/check_porting.py
MackieLoeffel, line 1: How about moving the scripts to a top-level scripts directory? This would allow us to get rid of the `IrisLean/src/` directory and reduce the directory nesting depth. (Maybe the `DumpPortingData.lean` file needs to be moved to the Iris folder in this case.)

IrisLean/LICENSE
MackieLoeffel, line 1: Should the license stay at the toplevel? I guess all packages will use the same license?

.github/workflows/build.yml
```diff
@@ -16,32 +16,55 @@ permissions:
...
+    steps:
+      - uses: actions/checkout@v6
+      - uses: leanprover/lean-action@v1
+        with:
+          lake-package-directory: IrisLeanMath
+          use-mathlib-cache: true
 
+  report:
+    # Generate and upload the porting report Pages artifact.
+    # Only runs on master; failures here don't block PRs.
```
MackieLoeffel: Do we want to only run on master? It might be nice to check on PRs that they do not introduce stale aliases.

MackieLoeffel:
Nice PR! I tried installing IrisLean following the instructions and it seems to work!

I am only wondering if we can / should keep IrisLean at the top-level such that one does not need to specify `git.subDir` for IrisLean, but only for IrisLeanMath? This would make it clear that `IrisLean` is the main library provided by this repo and the other libraries are additions. Maybe this would allow for at least IrisLean to show up on reservoir? (Not sure about this though.)

IrisLean/LICENSE
markusdemedeiros: Moved

.github/workflows/build.yml
markusdemedeiros: Agreed that it should run all the time. I think we need some more automation to observe stale aliases but perhaps we can revisit this after.

MackieLoeffel: I guess we could have a job that runs nightly and checks the repo against the current master of Iris Rocq? Then we would immediately see if e.g. things are renamed in Rocq.

markusdemedeiros:
I'm not sure the subdirectory idea is good. Under this setup, we'd need a folder accessible under the root Iris-Lean directory that is not meant to be part of the Iris-Lean project--I don't think it's worth it to remove one line from dependency files. The KLR project at AWS had this kind of nested structure and I remember it causing all kinds of IDE problems.

MackieLoeffel:
I see, then let's keep it in two separate directories.

scripts/check_porting.py
markusdemedeiros: Right on, stripped off the src directory and moved scripts as you suggested. Thanks for the feedback!

PR #331, opened 2026-04-21
feat: Add HeapLang syntax with notation

MackieLoeffel:
## Description
This PR adds the syntax of HeapLang and corresponding notation. Note that the notation has not been tested deeply, so there might be small errors in the levels.

## Checklist


Kaptch:
That's really cool! I will take a look and add some tests :)

Kaptch:
I've moved tests, and added some examples (simple programs and some programs from the Rocq version). I think deref should have a bit higher priority (e.g., to allow case: !loc).

I haven't noticed any problems while writing examples, so I think we can merge it. Would it be okay to keep higher precedence for dereference, or you had some example in mind, where it's undesirable?

It is nice to have the HeapLang early on, so that Fernando has some instance to test the language interface :)

Kaptch (approved):

MackieLoeffel:
Thanks for the review and adding the examples! I agree that 100 is a better precedence for load. In general, I am not fully sure about all the precedences. I will merge this PR and we can in the future iterate on the precedences if we notice anything weird.

PR #336, opened 2026-04-22
fix: fail CI build on warnings

MackieLoeffel:
Currently the build triggers a warning, but the CI does not catch this. This PR fixes the warning and enforces that CI checks for warnings.


markusdemedeiros:
Good call

PR #337, opened 2026-04-22
feat: Allow tactics in IPM synthesis

MackieLoeffel:
## Description
This PR adds the `ipm_tactic_instance` attribute that allows one to extend the IPM typeclass synthesis with custom tactics. These tactics can be mixed with normal instances and can recursively call typeclass synthesis again (see the tests for an example). This feature will be useful to implement `iframe` among other.

### DIFF
--- a/Iris/Iris/ProofMode/SynthInstance.lean
+++ b/Iris/Iris/ProofMode/SynthInstance.lean
@@ -7,6 +7,7 @@ module
 
 public import Qq
 public import Iris.BI
+public import Iris.ProofMode.SynthInstanceAttr
 
 public meta section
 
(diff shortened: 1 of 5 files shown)


lzy0505:
Very cool! I will try to have a look over the weekend.

lzy0505:
@MackieLoeffel Was this PR closed by accident or on purpose? Should I have a look?

MackieLoeffel:
I guess this was closed when I pushed to master directly on accident. It would be great if you could have a look.

Iris/Iris/Tests/Instances.lean
lzy0505, line 82: Irrelevant to this PR, but I found it strange that the same BI is synthesised twice here.
I don't think the one in the IPM trace is necessary, or at least it should hit the cache?

lzy0505 (approved):

Iris/Iris/Tests/Instances.lean
MackieLoeffel: I am also not fully sure what is happening here, but in general we have a lot of `[BI PROP]` assumptions in our lemmas and also in notation. I think this causes multiple searches for `[BI PROP]`. Not sure if we can do something about this. The cache is not shared between different instances of TC synth as far as I know, so I don't think it can help here.

MackieLoeffel:
Thanks for the review and the tweaks!

Iris/Iris/ProofMode/SynthInstance.lean
```diff
@@ -90,10 +56,42 @@ partial def synthInstanceMainCore (mvar : Expr) : MetaM (Option Unit) := do
...
+
+    -- first tactics and then instances. We cannot interleave them
+    -- since we don't know the priorities of the instances.
+    let tactics ← forallTelescopeReducing mvarType fun _ type => do
+      (synthTacticExt.getState (← getEnv)).getUnify type
+    let tactics := tactics.insertionSort fun e₁ e₂ => e₁.prio < e₂.prio
+    trace[Meta.synthInstance.tactics] m!"{tactics}"
+
+    let mctx ← getMCtx
+    for tac in tactics.reverse do
```
lzy0505, line 68: I thought I left a comment here. Can we just sort the list in descending order to avoid the reverse?

Iris/Iris/Tests/Instances.lean
lzy0505: Looked into this a bit, the standalone BI search seems to be a one-time thing. All searches in IPM share a cache, which I think is good enough.

PR #369, opened 2026-04-27
refactor: prefer Function.Injective over custom typeclass

ayhon:
## Description
I noticed we recently introduced an `Iris.Std.Injective` typeclass which we can replace with the builtin `Function.Injective` typeclass, since the definitions seem to be pretty much equivalent. I have made this PR since I believe it's better to use the built-in definitions as much as possible.

## Checklist


markusdemedeiros:
Good catch!

PR #371, opened 2026-04-27
fix: Change WSat to use identity OFunctor rather than constant 

markusdemedeiros:
Ran into this issue when trying to close off a fully adequacy proof in Approxis: It seems like in Rocq it's not a problem to use `constOF (IProp GF)` in the definition of the `Wsat` functor, but this caused me problems when trying to build a concrete model since then `GF` had to be recursive. To be honest I assumed the functor would be `Id` anyways. 

Not sure exactly how we should go here: the version in this PR deviates from Rocq but seems to work nicely with our existing infrastructure. Alternatively, we could think harder about why the Rocq version works and try to tweak `ElemG` etc. to match. 

Any thoughts about this? Paging @Kaptch specifically

### DIFF
--- a/Iris/Iris/Algebra/OFE.lean
+++ b/Iris/Iris/Algebra/OFE.lean
@@ -693,6 +693,15 @@ instance oFunctorConstOF [OFE B] : OFunctor (constOF B) where
 instance OFunctor.constOF_contractive [OFE B] : OFunctorContractive (constOF B) where
   map_contractive.1 := by simp [map]
 
+abbrev IdOF : COFE.OFunctorPre := fun _ B _ _ => B
+
+instance IdOF_OFunctor : COFE.OFunctor IdOF where
+  cofe := inferInstance
+  map _ g := g
+  map_ne := ⟨fun _ _ _ _ _ _ hg => hg⟩
+  map_id _ := .rfl
+  map_comp _ _ _ _ _ := .rfl
+
 end COFE
 
 /- Leibniz OFE structure on a type -/
@@ -1051,6 +1060,14 @@ instance [OFunctorContractive F] : OFunctorContractive (LaterOF F) where
   map_contractive.1 H z m := Dist.lt <| by
     have := (OFunctorContractive.map_contractive (F := F)).distLater_dist H
     simp_all only [Dist, DistLater, Function.uncurry, OFunctor.map, laterMap]
+
+instance laterIdOF_contractive : COFE.OFunctorContractive (LaterOF IdOF) where
+    map_contractive.1 := by
+      intro n x y H z
+      simp only [Function.uncurry, COFE.OFunctor.map, laterMap, IdOF]
+      show OFE.DistLater _ _ _
+      intros k Hk
+      exact (H k Hk).2 _
 
 end LaterOF
 
--- a/Iris/Iris/Examples.lean
+++ b/Iris/Iris/Examples.lean
@@ -5,3 +5,4 @@ public import Iris.Examples.Namesets
 public import Iris.Examples.Proofs
 public import Iris.Examples.Resources
 public import Iris.Examples.HeapLang
+public import Iris.Examples.ClosedProofs
--- a/Iris/Iris/Examples/ClosedProofs.lean
+++ b/Iris/Iris/Examples/ClosedProofs.lean
@@ -1,0 +1,78 @@
+/-
+Copyright (c) 2026 Sergei Stepanenko. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+module
+
+public import Iris.ProofMode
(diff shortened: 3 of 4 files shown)


Kaptch:
Sorry, I think I miss something. Could you please point out to the file or line where it breaks in Approxis with the old definition?

markusdemedeiros:
Hm. So I actually misread `wsat` a little bit. I was using `InvMapF` (a field of Rocq's `wsatGpreS`) instead of Rocq's `wsatΣ`, which didn't exist. 

However this leaves me even more confused about why Iris-Rocq uses the definition 
```coq
    wsatGpreS_inv : inG Σ (gmap_viewR positive (agreeR $ laterO (iPropO Σ)));
```
instead of 
```coq
    wsatGpreS_inv : inG Σ (gmap_viewR positive (agreeR $ LaterO IdOF));
```

My issue arises when trying to actually supply a list of `BundledGFunctors`. In Approxis, I had: 

```
noncomputable def otpSigma : BundledGFunctors := fun n =>
  match n with
  | 0  => ⟨InvMapF, by infer_instance⟩
  -- Other cases are irrelevant to this example

instance otpSigma_WsatGpreS : WsatGpreS otpSigma where
  inv := { τ := 0, transp := by unfold otpSigma; rfl } -- Prove the ElemG instance
  enabled := { τ := 1, transp := by unfold otpSigma; rfl }
  disabled := { τ := 2, transp := by unfold otpSigma; rfl }
```
This code works under the current PR where `InvMapF` is defined using the identity OFunctor instead of the constant `IProp` OFunctor. I briefly pushed a commit labeled `experiment` to this branch where I tried to mimic the `wsatGpreS`/`wsatΣ` of Rocq, though I'm reverting it to my initial PR for now, because it makes the `ElemG` instance into some really nasty dependent type thing instead of `rfl`. 

I guess I'm wondering a couple things. 
- Do you have any idea why Rocq uses `(iPropO Σ)` instead of `IdOF` like proposed in this PR? 
- Have you done a proof using invariants yet where you supply an explicit list of `OFunctors` like I've done in Approxis?
- If we were to add an analogue `wsatΣ`, do you know of an easy way to prove the `elemG` instance (given that this way goes through by `rfl`).

Apologies if I'm not being clear, I'm asking because you're a lot more of an expert in this part of the development than I am!

Kaptch:
Thanks for the example! I will take a look and try to see what we can do here. I think the issue is: we are are just instantiating a functor (with iProp), and in the Rocq version there are two things: functor that just gives shape, and the version that is supposed to be used, when it's solved. I think your approach is probably correct, and we should use a slightly different approach, when we specify ElemG. I'll read ElemG more carefully and experiment tomorrow to see if there's a good way to do it.
Probably, a temporary fix is to use fold/unfold lemmas, to reason about the functor and its instantiated version.

markusdemedeiros:
Nice, thanks! I'll keep this branch open so I can depend on it for Approxis. 

I think we should add a file to the examples folder which does this instantiation for invariants specifically (and does the full allocation/BundledGFunctors construction), as long as there's some way for that to work I think downstream logics like Approxis will be OK.

MackieLoeffel:
I don't know the technical details of this discussion, but I think it would be great if at the end there would be something like [this documentation in Iris](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/docs/resource_algebras.md?ref_type=heads) for how to obtain a closed proof.

Iris/Iris/Algebra/OFE.lean
```diff
@@ -1052,6 +1061,14 @@ instance [OFunctorContractive F] : OFunctorContractive (LaterOF F) where
     have := (OFunctorContractive.map_contractive (F := F)).distLater_dist H
     simp_all only [Dist, DistLater, Function.uncurry, OFunctor.map, laterMap]
```
Kaptch, line 1063: I think most of these functors and instances are in the leibniz-preserve PR. We probably should make a pass later. I think all the functors from ofe.v are there, but it's better to double check.

markusdemedeiros: Coincidentally, I just checked it out for review 😉

Kaptch:
Taking a look at it. I'll write a simple example of a closed proof obtained through iprop instance of logic with some resource (probably with wsat), to localize the problem, and see how to proceed from there (it also would be nice to have an example to add it to the readme).

markusdemedeiros:
@Kaptch Mystery solved: the difference is ultimately due to the fact that Iris-Lean's `ElemG` takes a functor as argument, whereas Rocq's `inG` takes a CMRA (we ended up not doing this to work better with typeclasses and avoid more bundling). I think this modified version is the right move for Iris-Lean.

Kaptch:
Thanks! That explains it, I will do the same for other circular resources, when porting.

PR #372, opened 2026-04-29
More HeapLang notation

suhr:
## Description

Add some notation used in iris-tutorial. ~~Also change `case: $e` syntax to `case $e with`.~~

The `case:` notation is replaced with a notation that more closely matches Rocq Iris notation and Lean syntax.

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/Examples/HeapLang.lean
+++ b/Iris/Iris/Examples/HeapLang.lean
@@ -48,27 +48,27 @@ def minMax : Exp :=
 
 def optionMap : Exp :=
   hl(λ f opt,
-    case: opt
-    | _ => injl(#())
-    | x => injr(f x))
+    match opt with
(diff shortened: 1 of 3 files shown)


suhr:
Actually, let me change that `case:` syntax once again...

suhr:
This should work. The order of variants is strict but this could be improved later.

Kaptch:
Thank you for your PR! I think inl/inr in patterns is a strict improvement for readability. For the case/match diff, I agree that it might be better to keep it relatively aligned with the Rocq syntax for the purpose of reusing tutorial materials. @MackieLoeffel what do you think?

MackieLoeffel:
Thanks for the PR! I think the new notation looks good. I did some clean up and added some tests.

PR #378, opened 2026-05-01
feat: iframe tactic and input/output handling for IPM synthesis

### LINKED ISSUE #358: Port iFrame

MackieLoeffel:
## Description

Implement the iframe tactic and make it available in specialization patterns and cases patterns.

Additional implement input / output handling to IPM synthesis roughly in the style of Hint Mode in Rocq. See the comment on `ParamKind` for a detailed description:
```
The parameters of a class declared with `ipm_class` are categorized into the following categories:

1. in: This is the default for a parameter when not annotated in another way.
2. out: These are parameters marked with `outParam`. These parameters must be mvars when starting typeclass search.
3. semiOut: These are parameters marked with `semiOutParam`. The preceding parameter must be of type `InOut`. The value of the `InOut` parameter decides whether the `semiOut` parameter is treated as an input or an output.
4. uncheckedIn: These are parameters marked with `uncheckedInParam`. These behave like `in` parameters, but allow mvars to match terms (see below).
5. inout: Parameters of type `InOut`, must appear before semiOut parameters (see above).

The following constraints apply to the parameters: (In the following, semiOut parameters are treated as inputs if the preceding `InOut` is `in` and as outputs if the `InOut` is `out`.)
1. semiOut parameters must have a preceding `InOut` parameter.
2. For each synthesis problem, all output parameters must be mvars.
3. When an input parameter is a mvar, it is not considered to match an instance which does not have an mvar at the top-level. This is to prevent accidentally instantiating mvars. Note that this only applies for mvars the top-level (matching the behavior of Hint Mode : ! in Rocq), since this is the simplest version to implement and catches most (all?) of the cases.
```

## Checklist

### DIFF
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -18,27 +18,18 @@ open Iris.BI
   that have pure preconditions (such as [ElimModal]) -/
 inductive PMError (msg : String) : Prop
 
-/-- [InOut] is used to dynamically determine whether a type class
-parameter is an input or an output. This is important for classes that
-are used with multiple modings, e.g., IntoWand. Instances can match on
-the InOut parameter to avoid accidentially instantiating outputs if
-matching on an input was intended. -/
-inductive InOut where
-  | in
-  | out
-
 inductive AsEmpValid.Direction where
   | into
   | from
 
 @[ipm_class]
-class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) {PROP : semiOutParam (Type _)} (P : outParam PROP) [semiOutParam (BI PROP)] where
+class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) (_ : InOut) (PROP : semiOutParam (Type _)) (_ : InOut) (bi : semiOutParam (BI PROP)) (P : outParam PROP) where
   as_emp_valid : (d = .into → φ → ⊢ P) ∧ (d = .from → (⊢ P) → φ)
 
-theorem asEmpValid_1 [BI PROP] (P : PROP) [AsEmpValid .into φ P] : φ → ⊢ P :=
-  AsEmpValid.as_emp_valid.1 rfl
-theorem asEmpValid_2 [BI PROP] (φ : Prop) [AsEmpValid .from φ (P : PROP)] : (⊢ P) → φ :=
-  AsEmpValid.as_emp_valid.2 rfl
+theorem asEmpValid_1 [bi : BI PROP] (P : PROP) [AsEmpValid .into φ .in PROP .in bi P] : φ → ⊢ P :=
+  (AsEmpValid.as_emp_valid .in .in).1 rfl
+theorem asEmpValid_2 [bi : BI PROP] (φ : Prop) [AsEmpValid .from φ .out PROP .out bi P] : (⊢ P) → φ :=
+  (AsEmpValid.as_emp_valid .out .out).2 rfl
 
 /- Depending on the use case, type classes with the prefix `From` or `Into` are used. Type classes
 with the prefix `From` are used to generate one or more propositions *from* which the original
@@ -52,7 +43,7 @@ class FromImp [BI PROP] (P : PROP) (Q1 Q2 : outParam PROP) where
 export FromImp (from_imp)
 
 @[ipm_class]
-class FromWand [BI PROP] (P : PROP) (Q1 Q2 : outParam PROP) where
+class FromWand [BI PROP] (P : PROP) (_ : InOut) (Q1 : semiOutParam PROP) (Q2 : outParam PROP) where
   from_wand : (Q1 -∗ Q2) ⊢ P
 export FromWand (from_wand)
 
@@ -140,7 +131,7 @@ class IntoPure [BI PROP] (P : PROP) (φ : outParam Prop) where
 export IntoPure (into_pure)
 
 @[ipm_class]
-class FromPure [BI PROP] (a : outParam Bool) (P : PROP) (φ : outParam Prop) where
+class FromPure [BI PROP] (a : outParam Bool) (P : PROP) (_ : InOut) (φ : semiOutParam Prop) where
   from_pure : <affine>?a ⌜φ⌝ ⊢ P
 export FromPure (from_pure)
 
@@ -154,9 +145,12 @@ class IntoExcept0 [BI PROP] (P : PROP) (Q : outParam PROP) where
   into_except0 : P ⊢ ◇ Q
 export IntoExcept0 (into_except0)
 
-/-- `FromModal` turns a goal `P : PROP2` into a modality `M : PROP1 → PROP2` applied to `Q : PROP1` under condition `φ`. -/
+/-- `FromModal` turns a goal `P : PROP2` into a modality `M : PROP1 → PROP2` applied to `Q : PROP1` under condition `φ`.
+
+`sel` is an input that can be provided by the user to match on the desired modality to introduce. It needs to be an `outParam` to make Lean happy since `PROP1` is an `outParam`. For the IPM TC synthesis, it needs to be an `uncheckedInParam` since it should match all modalities if the user provides an mvar.
+ -/
 @[ipm_class]
-class FromModal {PROP1 PROP2} [BI PROP1] [BI PROP2] (φ : outParam $ Prop) (M : outParam $ Modality PROP1 PROP2) (sel : semiOutParam PROP1) (P : PROP2) (Q : outParam $ PROP1) where
+class FromModal {PROP1 : outParam (Type _)} {PROP2} [outParam (BI PROP1)] [BI PROP2] (φ : outParam $ Prop) (M : outParam $ Modality PROP1 PROP2) (sel : outParam (uncheckedInParam PROP1)) (P : PROP2) (Q : outParam $ PROP1) where
   from_modal : φ → M.M Q ⊢ P
 export FromModal (from_modal)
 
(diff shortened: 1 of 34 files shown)


lzy0505:
Amazing work! I’ll review it this week.

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
```diff
@@ -1,28 +1,76 @@
...
 syntax ident : specPat
 syntax "%" term:max : specPat
-syntax "[" ident* "]" optional(" as " ident) : specPat
+syntax "[" frameIdent* "]" optional(" as " ident) : specPat
+syntax "[" "-" frameIdent* "]" optional(" as " ident) : specPat
+syntax "[>" frameIdent* "]" optional(" as " ident) : specPat
+syntax "[>" "-" frameIdent* "]" optional(" as " ident) : specPat
+syntax "[#" frameIdent* "]" optional(" as " ident) : specPat
+syntax "[" "$" "]" : specPat
+syntax "[#" "$" "]" : specPat
```
lzy0505, line 32: `[> $]` is missing?

MackieLoeffel:
@lzy0505 When do you think you will be able to have a look at this? I think it would be great to get this PR in relatively soon if possible since basically all other PRs to the proof mode are blocked on the changes in this PR.

lzy0505:
@MackieLoeffel Working on this now. Should be able to finish by the end of today. Sorry for the delay.

MackieLoeffel:
No worries, thank you for looking into this!

Iris/Iris/ProofMode/SynthInstanceAttr.lean
```diff
@@ -9,24 +9,172 @@ public import Qq
...
+def isUncheckedInParam (e : Expr) : Bool :=
+  e.isAppOfArity ``uncheckedInParam 1 || (e.isOutParam && e.appArg!.isAppOfArity ``uncheckedInParam 1)
+
+/--
+The parameters of a class declared with `ipm_class` are categorized into the following categories:
+
+1. in: This is the default for a parameter when not annotated in another way.
+2. out: These are parameters marked with `outParam`. These parameters must be mvars when starting typeclass search.
+3. semiOut: These are parameters marked with `semiOutParam`. The preceding parameter must be of type `InOut`. The value of the `InOut` parameter decides whether the `semiOut` parameter is treated as an input or an output.
+4. uncheckedIn: These are parameters marked with `uncheckedInParam`. These behave like `in` parameters, but allow mvars to match terms (see below).
```
lzy0505: I feel like using Lean's `semiOutParam` and `outParam` as markers is a bit confusing here.
Because `semiOutParam` has very different meaning in Lean's standard synthesis algorithm.

But I also feel like it is intended, if IIUC: it makes the fallback to Lean's algorithm work. 
Is it true? I don't know whether there is a better solution.

Iris/Iris/ProofMode/SynthInstance.lean
```diff
@@ -28,6 +34,9 @@ the IPM synthesis needs to be explicitly invoked via the functions in this file.
 The `ipm_backtrack` attribute on an instance tells the IPM synthesis to backtrack if instance instance can be applied, but
 its preconditions fail to synthesize. This is not enabled by default to avoid accidental exponential blow-ups.
 
+The `ipm_tactic_instance` attribute on a function of type `SynthTactic` declares a tactic that is used to solve synthesis problems for
+a given pattern. These tactics can call ipm synthesis recursively. See Tests/Instances.lean for examples.
+
 The `InOut` type in Classes.lean is used to dynamically determine, which parameters are inputs and which are outputs. IPM synthesis
 ignores `outParam` and `semiOutParam` annotations, but it is still recommended to add these annotations as documentation.
```
lzy0505, line 43: I removed the comments on InOut here, as I think those are outdated now.

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -140,7 +131,7 @@ class IntoPure [BI PROP] (P : PROP) (φ : outParam Prop) where
 export IntoPure (into_pure)
 
 @[ipm_class]
-class FromPure [BI PROP] (a : outParam Bool) (P : PROP) (φ : outParam Prop) where
+class FromPure [BI PROP] (a : outParam Bool) (P : PROP) (_ : InOut) (φ : semiOutParam Prop) where
```
lzy0505: Per point 2 of the docstring of ParamKind, `a` must be an mvar as it is marked with `outParam`. But there are places where it is actually provided. 
The following example can trigger the bug.
```
example [BI PROP] [BIAffine PROP] (φ1 φ2 : Prop) (Q : PROP) :
    (⌜φ1⌝ → ⌜φ2⌝ : PROP) ⊢ Q -∗ Q := by
  iintro Himp
  iintro HQ
  ipure Himp
  iexact HQ
```
I fixed it by marking `a` as a `semiOutParam`.

lzy0505 (commented):
This is an excellent PR.

Regarding the iframe implementation, I haven’t checked if there are more missing features than those marked with TODO but the existing code looks great!

The additional in/out part took me some time. In general, I like the idea and think the code can be merged. However, I’m not sure if the implementation approach is the best. Currently, checking an IPM class’s markings requires examining all potential invocations of the IPM synthesis algorithm for the class. I tried to do this for a few classes while understanding the code and believe I found a mistake in the `FromPure` marking.  See my inline comments. I didn’t have time to check all but I think such bugs can be caught, reported, and fixed as more IPM proofs are developed. I don't know how hard it is, but I feel like a more static approach is desirable. 
I also feel like this part of the code needs more comments, and perhaps some renaming to avoid confusion (e.g. my inline comments).

In the commit I pushed, I made many stylistic changes, added some comments, added some tests, fixed the `FromPure` bug and corrected an emoji display issue in synthInstance traces. Feel free to revert/edit any of my changes.

lzy0505 (approved):

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -38,147 +41,153 @@ proposition can be derived. Type classes with the prefix `Into` are used to gene
 used to indicate that certain propositions should be intuitionistic. -/
 
 @[ipm_class]
-class FromImp [BI PROP] (P : PROP) (Q1 Q2 : outParam PROP) where
+class FromImp {PROP} [BI PROP] (P : PROP) (Q1 Q2 : outParam $ PROP) where
```
MackieLoeffel, line 44: @lzy0505 Do you have a strong opinion on using `outParam $ PROP` instead of `outParam PROP`? The seem equivalent to me and I am wondering why to use `$` if it is not necessary.

lzy0505: Previously, it was a bit randomly used, so I tried to make it more consistent for the markers. 
I'm slightly in favour of using the dollar sign everywhere because it makes it more obvious that the thing on the left is a marker. Feel free to revert if you don't like.

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
MackieLoeffel: It will be added by https://github.com/leanprover-community/iris-lean/pull/379/changes/4a8bd2c26fe1c9f3979c8b2c13d37a01b3112457#diff-0cab7e3ee3b99ecf00bc6bbc7e1c81006adf7257321e1f5ab751888d9b9f94ca

Iris/Iris/ProofMode/SynthInstanceAttr.lean
MackieLoeffel: We are still using Lean's synthesis order checking and for this is is useful to use `semiOutParam` and `outParam`. (The synthesis order checking is not perfect for IPM classes, but I still find it useful.) I will add a comment about this.

Iris/Iris/ProofMode/Classes.lean
MackieLoeffel: Indeed, there was a problem with `FromPure`, but I don't think making `a` a semiOutParam is the right solution. Instead, I tweaked the instances such that they work with `outParam`.

Iris/Iris/ProofMode/Classes.lean
MackieLoeffel: I see. I don't have a strong opinion and at least your version is consistent, so we can stick with that.

Iris/Iris/ProofMode/Classes.lean
lzy0505: Can you elaborate on why it is not?

MackieLoeffel: The version with the semiOutParam fails the following tests:
```
/-- Tests `ipure` with implication containing pure -/
example [BI PROP] (Q : PROP) : <affine> (⌜φ1⌝ ∧ ⌜φ2⌝ → ⌜φ3⌝)  ⊢ Q -∗ Q := by
  iintro Hφ
  iintro HQ
  ipure Hφ
  iexact HQ

/-- Tests `ipure_intro` with wand containing pure and affine lhs -/
example [BI PROP] : ⊢@{PROP} (<affine> ⌜φ2⌝ -∗ emp) := by
  ipure_intro
  intro _; trivial

/-- Tests `ipure_intro` with wand containing pure and absorbing rhs -/
example [BI PROP] : ⊢@{PROP} (⌜φ2⌝ -∗ <absorb> emp) := by
  ipure_intro
  intro _; trivial
```

This is because instances like
```
instance fromPure_pure_and (a1 a2 : Bool) (φ1 φ2 : Prop) [BI PROP] (P1 P2 : PROP)
    [h1 : FromPure .out a1 P1 io φ1] [h2 : FromPure .out a2 P2 io φ2] :
    FromPure ioA (a1 || a2) iprop(P1 ∧ P2) io (φ1 ∧ φ2) where
```
don't work for the `ioA = .in` case where the input is `false` because `false` and `(a1 || a2)` don't match syntactically and thus the instance is pruned by the discrimination tree. There were also some other instances I had to fix. I am currently waiting on Claude to do the proofs on the new instances and then I can push the changes.

MackieLoeffel:
Thanks for the review and all the cleanup!

> Currently, checking an IPM class’s markings requires examining all potential invocations of the IPM synthesis algorithm for the class. [...] I don't know how hard it is, but I feel like a more static approach is desirable.

I agree that the current approach can still lead to bugs where the input / output moding of an instance is incorrect, but it is only detected when one tries to use the instances. Having a static checker for this would be great, but this would be some work and probably it is best to do this in a separate PR.

Iris/Iris/ProofMode/Classes.lean
MackieLoeffel: The changes are in https://github.com/leanprover-community/iris-lean/pull/378/changes/05b1a38fa7ccef58aad3d07337bec53141ad749d

lzy0505: I see. Thanks for the explanation!

lzy0505:
@MackieLoeffel Yes, I agree such a static checker should belong to a separate PR. I think this one is ready for merging!

MackieLoeffel:
Great, thanks again for the review! Merging...

PR #380, opened 2026-05-04
feat: Simplify definition of le_upd

MackieLoeffel:
The P is constant throughout the recursion, so we might as well take a fixpoint at iProp instead of iProp → iProp.

Mirrors https://gitlab.mpi-sws.org/iris/iris/-/merge_requests/1215


Kaptch:
Thanks!

PR #382, opened 2026-05-07
feat: distinguish uniq from name in Hyps

ayhon:
## Description
Differentiate the types of `uniq` and `name` fields in `Hyps.hyp`. Mimics how `FVarId` is implemented.

## Checklist
* [X] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [X] I have updated `PORTING.md` as appropriate
* [ ] I have added my name to the `authors` section of any appropriate files

PR #387, opened 2026-05-11
feat: iloeb

### LINKED ISSUE #361: Port iLöb

ayhon:
## Description
Implements the `iLöb` tactic.

Addresses #361

## Checklist
* [X] I have updated `PORTING.md` as appropriate

### DIFF
--- a/Iris/Iris/ProofMode/Patterns/SelPattern.lean
+++ b/Iris/Iris/ProofMode/Patterns/SelPattern.lean
@@ -52,8 +52,8 @@ partial def SelPat.parse (pats : TSyntaxArray `selPat) : MacroM (List SelPat) :=
 public meta section
 
 inductive SelId where
-| lean (id : FVarId)
-| pm (ivar : IVarId)
+| pure (id : FVarId)
+| ipm (ivar : IVarId)
 deriving BEq, Hashable, Repr
 
 @[rocq_alias esel_pat]
@@ -62,17 +62,17 @@ structure SelTarget where
   /- Was this target specified explicitly or is it from a glob like ∗? -/
   explicit : Bool
 
-/-- Resolve selection patterns to concrete proofmode hypotheses (`.pm`) and Lean locals (`.lean`). -/
+/-- Resolve selection patterns to concrete proofmode hypotheses (`.ipm`) and Lean locals (`.lean`). -/
 def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget)
   | .ident name =>
-      return [⟨.pm (← hyps.findWithInfo name), true⟩]
+      return [⟨.ipm (← hyps.findWithInfo name), true⟩]
   | .leanIdent name => do
       let ldecl ← getLocalDeclFromUserName name.getId
-      return [⟨.lean ldecl.fvarId, true⟩]
+      return [⟨.pure ldecl.fvarId, true⟩]
   | .intuitionistic =>
-      return hyps.intuitionisticIVarIds.map (⟨.pm ·, false⟩)
+      return hyps.intuitionisticIVarIds.map (⟨.ipm ·, false⟩)
   | .spatial =>
-      return hyps.spatialIVarIds.map (⟨.pm ·, false⟩)
+      return hyps.spatialIVarIds.map (⟨.ipm ·, false⟩)
   | .pure => do
       -- `%` selects user-facing Lean pure assumptions, so we keep only `Prop` hypotheses.
       let mut hyps := #[]
@@ -81,7 +81,7 @@ def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget
           continue
         if ! (← isProp ldecl.type) then
           continue
-        hyps := hyps.push (⟨.lean ldecl.fvarId, false⟩)
+        hyps := hyps.push (⟨.pure ldecl.fvarId, false⟩)
       return hyps.toList
 
 def SelPat.resolve (hyps : Hyps bi e) (pats : List SelPat) :
--- a/Iris/Iris/ProofMode/Patterns/SelPattern.lean
+++ b/Iris/Iris/ProofMode/Patterns/SelPattern.lean
@@ -53,7 +53,7 @@ public meta section
 
 inductive SelId where
 | lean (id : FVarId)
-| pm (uniq : IVarId)
+| pm (ivar : IVarId)
 deriving BEq, Hashable, Repr
 
 @[rocq_alias esel_pat]
--- a/Iris/Iris/ProofMode/Patterns/SelPattern.lean
+++ b/Iris/Iris/ProofMode/Patterns/SelPattern.lean
@@ -61,18 +61,20 @@ structure SelTarget where
   target : SelId
   /- Was this target specified explicitly or is it from a glob like ∗? -/
   explicit : Bool
+  persistent? : Bool
 
 /-- Resolve selection patterns to concrete proofmode hypotheses (`.ipm`) and Lean locals (`.lean`). -/
 def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget)
-  | .ident name =>
-      return [⟨.ipm (← hyps.findWithInfo name), true⟩]
+  | .ident name => do
+      let ⟨ivar, persistent?⟩ ← hyps.findWithInfoPersistent name
+      return [⟨.ipm ivar, true, persistent?⟩]
   | .leanIdent name => do
       let ldecl ← getLocalDeclFromUserName name.getId
-      return [⟨.pure ldecl.fvarId, true⟩]
+      return [⟨.pure ldecl.fvarId, true, true⟩]
   | .intuitionistic =>
-      return hyps.intuitionisticIVarIds.map (⟨.ipm ·, false⟩)
+      return hyps.intuitionisticIVarIds.map (⟨.ipm ·, false, true⟩)
   | .spatial =>
-      return hyps.spatialIVarIds.map (⟨.ipm ·, false⟩)
+      return hyps.spatialIVarIds.map (⟨.ipm ·, false, false⟩)
   | .pure => do
       -- `%` selects user-facing Lean pure assumptions, so we keep only `Prop` hypotheses.
       let mut hyps := #[]
@@ -81,7 +83,7 @@ def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget
           continue
         if ! (← isProp ldecl.type) then
           continue
-        hyps := hyps.push (⟨.pure ldecl.fvarId, false⟩)
+        hyps := hyps.push (⟨.pure ldecl.fvarId, false, true⟩)
       return hyps.toList
 
 def SelPat.resolve (hyps : Hyps bi e) (pats : List SelPat) :
--- a/Iris/Iris/ProofMode/Patterns/SelPattern.lean
+++ b/Iris/Iris/ProofMode/Patterns/SelPattern.lean
@@ -51,23 +51,28 @@ partial def SelPat.parse (pats : TSyntaxArray `selPat) : MacroM (List SelPat) :=
 
 public meta section
 
+inductive SelId where
+| lean (id : FVarId)
+| pm (uniq : IVarId)
+deriving BEq, Hashable, Repr
+
 @[rocq_alias esel_pat]
 structure SelTarget where
-  target : IVarId ⊕ FVarId
+  target : SelId
   /- Was this target specified explicitly or is it from a glob like ∗? -/
   explicit : Bool
 
-/-- Resolve selection patterns to concrete proofmode hypotheses (`.inl`) and Lean locals (`.inr`). -/
+/-- Resolve selection patterns to concrete proofmode hypotheses (`.pm`) and Lean locals (`.lean`). -/
 def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget)
   | .ident name =>
-      return [⟨.inl (← hyps.findWithInfo name), true⟩]
+      return [⟨.pm (← hyps.findWithInfo name), true⟩]
   | .leanIdent name => do
       let ldecl ← getLocalDeclFromUserName name.getId
-      return [⟨.inr ldecl.fvarId, true⟩]
+      return [⟨.lean ldecl.fvarId, true⟩]
   | .intuitionistic =>
-      return hyps.intuitionisticIVarIds.map (⟨.inl ·, false⟩)
+      return hyps.intuitionisticIVarIds.map (⟨.pm ·, false⟩)
   | .spatial =>
-      return hyps.spatialIVarIds.map (⟨.inl ·, false⟩)
+      return hyps.spatialIVarIds.map (⟨.pm ·, false⟩)
   | .pure => do
       -- `%` selects user-facing Lean pure assumptions, so we keep only `Prop` hypotheses.
       let mut hyps := #[]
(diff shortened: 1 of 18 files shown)


Iris/Iris/ProofMode/Tactics/Loeb.lean
```diff
@@ -0,0 +1,124 @@
...
+inductive RevertTarget where
+| lean (id : FVarId)
+| pm (persistent? : Bool) (uniq : IVarId)
+deriving BEq, Hashable, Repr
+
+def RevertTarget.toSelTarget : RevertTarget → SelTarget
+  | .lean id => .lean id
+  | .pm _ ivar => .pm ivar
+
+def ProofModeM.revertingTelescope {α} (mvar : MVarId)(g : IrisGoal)(hs : List RevertTarget) (k : MVarId → IrisGoal → ProofModeM (IrisGoal × α)) :
```
MackieLoeffel: This function does not follow the general pattern how proofmode tactics are written. Instead of directly manipulating `MVarId`s and `IrisGoal`s, proof mode tactics take a `Hyps` and a `goal` as an argument and produce a new `Hyps` and `goal` as the result. Each proofmode tactic only has a single `MVarId.assign` call. You can look at the other tactics, e.g. at `ispecialize` or `iintro` for how tactics are written. This should also make the functions `ProofModeM.getMainGoal` and similar unnecessary. If you have questions for how to rewrite this, let me know.

MackieLoeffel (changes requested):

Iris/Iris/ProofMode/Patterns/SelPattern.lean
```diff
@@ -46,19 +46,22 @@ partial def SelPat.parse (pats : TSyntaxArray `selPat) : MacroM (List SelPat) :=
 
 public meta section
 
-abbrev SelTarget := IVarId ⊕ FVarId
+inductive SelTarget where
+| lean (id : FVarId)
+| pm (uniq : IVarId)
```
MackieLoeffel: The naming convention for variables of type `IVarId` is `ivar`, not `uniq`. (I changed this in #382.)

ayhon: Indeed, I missed this in the rebase :pray:

Iris/Iris/ProofMode/Tactics/Loeb.lean
ayhon: I have indeed found this pattern in the code, but I wasn't sure how to best compose tactics with this interface. In particular, I have understood that `Hyps |- goal` usually corresponds to the main target, and that we return the assignment for this main target. However, I don't really see how this API can be used to compose different tactics. In the end, once I am done executing `iRevertCore` I need to continue the proof under the new generated `Hyps' |- goal'`, but to retrieve it I need to use either a `ProofModeM.getMainGoal`, or a continuation, which seems to be the method preferred by `iModIntros`. Do you suggest I rewrite this in terms of continuations, or am I missing something?

MackieLoeffel: You can compose the tactics by just calling the different `...Core` functions (though sometimes they do not have the `Core` postfix, naming is a bit inconsistent). See for example [here](https://github.com/leanprover-community/iris-lean/blob/f1ac197bc4ef0d48c42409fcf0a6ff8123115659/Iris/Iris/ProofMode/Tactics/Intro.lean#L91) where `iintros` calls `icases` or [here](https://github.com/leanprover-community/iris-lean/blob/f1ac197bc4ef0d48c42409fcf0a6ff8123115659/Iris/Iris/ProofMode/Tactics/Cases.lean#L206) where `icases` calls `iclear`. Note that none of these functions need something like `getMainGoal`. The problem in this case is that the `iRevertCore` function that this PR adds also does not follow the right style and thus it is tricky to call.
If you want, I can sketch the `iRevertCore` function with the right signature.

ayhon: But for some of these functions, like `iIntroCore` or `iRevertCore`, I would need to be able to get the context after the operation has finished. In these cases the `goal` is always the same, but `hyps` does change at each operation, so I need to retrieve it somehow. That's what I was using `getMainGoal` for.

I see how I could get around this by doing what `iModIntro` is doing, asking for a continuation to run on the new `hyps`-`goal` pair, which is what I meant by "rewrite this in terms of continuations". The other way would be to change the signature of these functions to also give me access to the resulting transformed hypothesis, such as `iSpecializeCore`, but I'm not a big fan of this way.

I have started work on rewriting my solution in the continuation style. For this, the signature of `iRevertCore` would look something like this:

```lean4
abbrev ProofModeTactic : Type _ := ∀ {u : Level}{prop: Q(Type $u)}{bi : Q(BI $prop)}{e : Q($prop)}
  (hyps : Hyps bi e)(goal: Q($prop)), ProofModeM Q($e ⊢ $goal)

def ProofModeM.revertingTelescope
  (hs : List RevertTarget)
  (k : ProofModeTactic) : ProofModeTactic := -- ...
-- `k` is run on the generalized environment.
```

Is something like this what you had in mind?

MackieLoeffel: Yes, this is roughly what I had in mind, but I am not sure if you actually need to pass a continuation in this case. Instead, it might be possible to just return the new hyps and goal. See for example [`iSpecializeCore`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/Tactics/Specialize.lean#L124), which works without a continuation. (The signature will be slightly different in your case since `iRevertCore` needs to return a new goal, not just change the Hyps.)

(Though thinking about this again, maybe you need a continuation to remove the fvars from the context for goals created in the continuation? Not sure about this.)

I would also not introduce an abstraction like `ProofModeTactic` at this point, but follow the style of the other functions that take a continuation like `iModIntro` or `iIntros`. Note that different proof mode functions have different signatures and not all tactics use continuation passing style. Having such an abstraction can be discussed in a separate issue.

ayhon: I have addressed the changes you mentioned. It took me more time than I would have liked since I had to debug some silly issues, but I'm quite satisfied with the final result. Lmk what you think of it and if it was closer to what you were expecting

MackieLoeffel: Great! I think the changed version looks very good! Thanks for working on this. I have minor stylistic comments, but it is probably the easiest if I make a pass over the code and directly apply them. I hope to find time for this in the next week.

Iris/Iris/ProofMode/Tactics/Apply.lean
```diff
@@ -59,8 +58,8 @@ elab "iapply" colGt pmt:pmTerm : tactic => do
     -- ensure the context can be discarded
     let LOption.some _ ← trySynthInstanceQ q(TCOr (Affine $e) (Absorbing $goal))
       | throwError "iapply: the context {e} is not affine and goal not absorbing"
-    have rfl : Q($e ∗ □?$p $out ⊣⊢ $e ∗ □?$p $out) := q(.rfl)
-    mvar.assign q($(pf).trans (assumption (Q := $goal) $(rfl)))
+    -- have rfl : Q($e ∗ □?$p $out ⊣⊢ $e ∗ □?$p $out) := q(.rfl)
+    mvar.assign q($(pf).trans (assumption (Q := $goal) .rfl)) -- TODO: Is this better?
```
MackieLoeffel: Seems better, but I would probably refactor this to have a more specialized lemma instead of `assumption` that does not need the `rfl` proof as the argument.

Iris/Iris/ProofMode/Patterns/SelPattern.lean
```diff
@@ -51,23 +51,28 @@ partial def SelPat.parse (pats : TSyntaxArray `selPat) : MacroM (List SelPat) :=
 
 public meta section
 
+inductive SelId where
+| lean (id : FVarId)
+| pm (ivar : IVarId)
```
MackieLoeffel: I would rename `pm` to `ipm`. We usually use `ipm` as the abbreviation to refer to the proof mode.

Iris/Iris/ProofMode/Patterns/SelPattern.lean
```diff
@@ -51,23 +51,28 @@ partial def SelPat.parse (pats : TSyntaxArray `selPat) : MacroM (List SelPat) :=
 
 public meta section
 
+inductive SelId where
+| lean (id : FVarId)
```
MackieLoeffel: maybe `pure` instead of `lean`?

Iris/Iris/ProofMode/Tactics/Intro.lean
```diff
@@ -52,43 +52,51 @@ theorem wand_intro_spatial [BI PROP] {P Q A1 A2 : PROP}
...
+/--
+Introduce the hypothesis specified by `pats` into the context given by `P` (structured  as `hyps`).
+The type of the current goal is given by `Q`.
+
+This function returns the proof of `P ⊢ Q` to be assigned. The new context is included in the
+`goals` directly by the tactic.
+-/
+partial def iIntroCore {prop : Q(Type u)} {bi : Q(BI $prop)}
+  {P} (hyps : Hyps bi P) (Q : Q($prop)) (pats : List (Syntax × IntroPat))
+  (k : ∀ {u : Level}{prop: Q(Type $u)}{bi : Q(BI $prop)}{e : Q($prop)}(hyps : Hyps bi e)(goal: Q($prop)), ProofModeM Q($e ⊢ $goal) := addBIGoal) :
```
MackieLoeffel: The `k` does not need to take `u`, `prop`, and `bi` as arguments (also applies to other places).

Iris/Iris/ProofMode/Tactics/Intro.lean
```diff
@@ -52,43 +52,51 @@ theorem wand_intro_spatial [BI PROP] {P Q A1 A2 : PROP}
...
     withRef ref do
-    iModIntroCore hyps Q (← `(_)) (iIntroCore · · pats)
+    iModIntroCore hyps Q (← `(_)) (iIntroCore · · pats k)
   | (ref, .intro (.pure n)) :: pats =>
     withRef ref do
     let v ← mkFreshLevelMVar
-    let α : Quoted q(Sort v) ← mkFreshExprMVarQ q(Sort v)
-    let Φ : Quoted q($α → $prop) ← mkFreshExprMVarQ q($α → $prop)
+    let α ← mkFreshExprMVarQ q(Sort v)     -- TODO: We used to have the types of these declarations explicitly.
+    let Φ ← mkFreshExprMVarQ q($α → $prop) --       But why?
```
MackieLoeffel: There were weird Qq bugs which required annotating the type at some point, but I guess this is no longer necessary.

Iris/Iris/ProofMode/Tactics/Loeb.lean
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Iris.ProofMode
+
+open Lean Meta Elab.Tactic Qq
+
+public meta section
+
+inductive RevertTarget where
+| lean (id : FVarId)
+| pm (persistent? : Bool) (ivar : IVarId)
```
MackieLoeffel: Instead of adding a new `RevertTarget`, I would add the `persistent` field to `SelTarget` and fill it during `SelPat.resolve`

Iris/Iris/ProofMode/Tactics/Loeb.lean
```diff
@@ -0,0 +1,93 @@
...
+  All spatial hypothesis are generalized in the induction hypothesis so that
+  this one can be included in the intuitionistic context.
+
+  Optionally, other variables can be generalized over through the
+  `generalizing selPat*` syntax.
+-/
+syntax (name := iloeb) "iloeb " " as " binderIdent (" generalizing " (ppSpace colGt selPat)+)? : tactic
+
+@[inherit_doc iloeb]
+elab_rules : tactic
```
MackieLoeffel: Usually we use `elab` instead of `syntax` + `elab_rule`. Is there a specific reason to use `syntax` + `elab_rule` here?

Iris/Iris/ProofMode/ProofModeM.lean
```diff
@@ -123,3 +134,25 @@ def ProofModeM.runTactic (x : MVarId → IrisGoal → ProofModeM α) (s : ProofM
 
   replaceMainGoal (nonDep ++ dep).toList
   return res
+
+def ProofModeM.pruneSolvedGoals : ProofModeM Unit := do
```
MackieLoeffel: These functions here are not used as far as I see.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2166,3 +2166,132 @@ example [BI PROP] [BIAffine PROP] (Q : Nat → PROP) : (Q 0 ⊢ ∃ x, False ∨
...
+  □IH : ▷ (<affine> ⌜H₁⌝ -∗ P -∗ Q)
+  ∗p : P
+  ⊢ Q
+-/
+#guard_msgs in
+example (H₁ : Prop) (P Q : PROP) :
+    H₁ → ⊢ P -∗ Q := by
+  iintro %h1 p
+  iloeb as IH generalizing %h1
+
```
MackieLoeffel, line 2289: It would be also good to have some failing test cases. E.g. what happens when one just writes `iloeb` (without `as IH`)? Rocq has a very confusing error there.

MackieLoeffel (commented):
Here are some small points I noticed when looking over this. You can fix this or I can take care of it when I make my pass.

Iris/Iris/ProofMode/Tactics/Loeb.lean
ayhon: I was just running into issues with the definition, and I find it easier to debug those when separating parsing and elaboration. Probably now those two can be merged, I simply forgot to test them

Iris/Iris/ProofMode/ProofModeM.lean
ayhon: Indeed, leftovers from my prior code, my bad

Iris/Iris/ProofMode/Tactics/Intro.lean
ayhon: Would it make sense to keep them then, in case they creep up again?

Iris/Iris/ProofMode/Tactics/Intro.lean
ayhon: Actually, because of `iModIntroCore`, they might have to generalize over at least `prop`. `iIntroCore` specifically needs it to compile, the only one that maybe gets away with sharing the same `u` `prop` and `bi` is `iRevertCore`. However, `u` is for sure shared everywhere, and I would expect it to not really have to change.

Iris/Iris/ProofMode/Tactics/Loeb.lean
ayhon: Ok, the issue I was having is matching under the `?` and `*`, I'm not sure how to do that in the `elab` syntax

Iris/Iris/Tests/Tactics.lean
ayhon: I believe this is a syntax error, and I'm not sure I can have a failing test case for it. I do have a test for the warning on generalizing over spatial hypothesis, saying "this is unnecessary, `iloeb` already generalizes over the whole spatial context".

I haven't found many instances where `iloeb` could fail, other than an `BILoeb` instance not being available. I've added this later test.

ayhon:
Addressed most review comments. Left a couple open for further comment. 

I can't reproduce the CI build failure locally (with all other `git` changes staged)

EDIT: I'm guessing they're due to warnings?

MackieLoeffel (approved):
I made a pass over the PR and it now looks good to me. @ayhon Let me know if you see any problem in my last commit. Otherwise, I will merge this in the next days. Thanks again for the PR!

Iris/Iris/ProofMode/Tactics/Loeb.lean
MackieLoeffel: I changed the code to use `elab` and a macro instead of `?`, which is how we handle this for other tactics..

ayhon:
I see no problem in your last commit, lgtm!

Maybe we should put the trace message of `iRevertIntro` under `trace.irevertintro` (`trace.iRevertIntro`?) instead of `trace.irevert` though

MackieLoeffel:
Great! I changed the traceclass to irevertintro and merged!

Issue #253, opened 2026-04-07
Port bi/weakestpre.v


lzy0505:
## Summary

Port [`bi/weakestpre.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/bi/weakestpre.v).

## Dependencies

**Rocq dependencies:**

- #242

Issue #390, opened 2026-05-12
Support existential quantifiers in iFrame


MackieLoeffel:
Originally implemented in Rocq in https://gitlab.mpi-sws.org/iris/iris/-/merge_requests/1017 and later tweaked by https://gitlab.mpi-sws.org/iris/iris/-/merge_requests/1035

PR #389, opened 2026-05-12
feat: fixpoint base of weakest precondition

lihaokun:
## Description

Lean 4 port of the fixpoint base of Coq Iris's program logic weakest
precondition (`iris/program_logic/weakestpre.v`). Adds `Stuckness`, the
`IrisGS` typeclass, `wp_pre` (with `num_laters_per_step` parameterization),
its `Contractive` instance, and `wp` / `wp_unfold` via the upstream
`fixpoint` / `fixpoint_unfold` machinery.

Builds on #370 (`Language` interface).

Contributes to porting `weakestpre.v` in `PORTING.md`.

### `Iris/Iris/ProgramLogic/WeakestPre.lean`

- `Stuckness` inductive — 1:1 with Coq `stuckness`
- `IrisGS` typeclass — all 5 fields of Coq `irisGS_gen`:
  - `iris_invGS` via `extends InvGS_gen hlc GF`
  - `state_interp` / `fork_post`
  - `num_laters_per_step` / `state_interp_mono`
- `wp_pre` (body 1:1 with Coq, using `={∅}▷=∗^[num_laters_per_step ns + 1]`)
- `wp_pre_contractive` Contractive instance
- `wp := fixpoint (wp_pre s)`
- `wp_unfold` via upstream `fixpoint_unfold`

Adds one private helper `step_fupdN_ne` to factor out the
`BIFUpdate.ne + later_ne + BIFUpdate.ne` induction over
`num_laters_per_step ns` needed by the Contractive proof.

### Adaptations vs Coq

- **`Stuckness` constructors lowercase**: `notStuck` / `maybeStuck` per Lean
  convention.
- **Plain function-space arrows**: `CoPset → Expr → (Val → IProp GF) → IProp GF`
  rather than Coq's explicit discrete-OFE arrows `-d>`. The upstream
  function-space COFE instance derives the equivalent OFE structure
  automatically (and the source types here have no non-trivial OFE structure).
- **`wp_pre_contractive` proof is manual**: ~25 lines plus the small private
  `step_fupdN_ne` helper, since iris-lean doesn't yet have the equivalent of
  Coq's `f_contractive` tactic. Structure follows
  `Iris/Examples/IProp.lean:wp_F_contractive`. Once tactic automation lands
  (cf. #387 for `iLöb`), this can likely be shortened.
- **`wp_unfold` proof**: wraps `wp_pre s` into a `ContractiveHom`, then
  combines `fixpoint_unfold` with pointwise evaluation on the function-space
[...]

PR #393, opened 2026-05-13
feat: add weakest precondition

ayhon:
## Description
Porting of the weakest precondition definitions (`program_logic/weakestpre.v`) and notation (`bi/weakestpre.v`).

Implements #253.

## Checklist


ayhon:
Things left to do:
 - [x] Assign priorities to the Proof Mode classes
 - [ ] Fix `WP` notation issues
    - Currently, we need to use `iprop(..)` in the postconditions
    - The term representing the expression needs a binding of `max`. This seems to be an issue with `@` confusing the parser. Consider changing the notation or doing something more clever in the `syntax` 
    - The `WP` notation doesn't seem to infer that the return value `v` in the postcondition `{{ v, Q }}` is always of the `Val` type. It can probably be fixed by inserting an explicit coercion somewhere, but that doesn't seem to be doable from the syntax. It might need an auxiliary definition.
 - [ ] All `TCEq (toVal e) None` were rewritten to plain `toVal e = none` preconditions. I conjecture that we don't need `TCEq` anymore, but this may actually be false.
 - [x] Some proof mode instances, such as `AddModal` or `ElimAcc`, were not added since the classes are not defined. We could also just define them and add the instances.
 - [x] Evaluate whether the `StateInterp` and `NumLatersPerStep` classes make sense. I separated these definitions preemptively, but they might not be useful in practice.
 - [x] Some proofs could be simplified. Look at #389 for `wp_pre_contractive` for instance, it seems to be a bit shorter than mine.

ayhon:
Since instances in [LaterCredits.lean](https://github.com/ayhon/iris-lean/blob/fele/feat/add-weakestpre/Iris/Iris/Instances/Lib/LaterCredits.lean#L417) don't seem to have priorities, I guess it's fine to not have them. It doesn't seem like they're checked by the custom `synthInstances` for `ipm_class`es.

Kaptch:
Hi! Thank you for the PR!
Would it be okay if I take a look and see what is left to do to merge it (e.g. the notation and TCeq points)?
I think for now AddModal and ElimAcc can be left aside, and probably added to the issue that targets them, so we don't forget about them in the future.

ayhon:
Sure! I think the main issue is fixing the notation a bit. There's some low hanging fruit (such as integrating it with `iprop(...)`), but some other points I find a bit more challenging, at least at my current level of understanding of how syntax works. I would like to relax the priority for the inner expression term, so one doesn't have to write parenthesis for a complex expression. I managed to get it working changing `@` for `@@`, but there might be a way to keep the original Rocq notation by using a judicious mix of `atomic(...)` and prioritie, which I think would be the most ideal.

About TCEq, I would like first to understand why it was needed in Rock to see whether it makes sense in Lean. That's why I purposefully didn't use it, so I could use this as an excuse to understand it's use case.

Another point, on the bi/weakestpre.v file I porpusefully didn't port texan triples since Ralf mentioned we might have a different plan for those. So texan triple syntax was technically out of scope for this PR, although I leave to your discretion if it makes sense to complete what's missing  of it and merge it for now.

markusdemedeiros:
Let me know when I can help review this--looking forward to getting it in asap!

Kaptch:
Hi again! Just in case, is it okay if this PR is merged without texan triples? I went over the BI part without them, but can restore if you want to keep some parts of texan triples in this PR (asking before pushing results of the current pass).

One question, what did you mean by "so one doesn't have to write parenthesis for a complex expression"? From my understanding, for expressions we probably always will have some sort of wrapper, similar to hl(..) for HeapLang.
In the printing direction, it should be solved by using a parenthesizer for hl_expr. 

@markusdemedeiros I will do the program logic part tomorrow, and send a message here, when it's ready from my side. Would it be okay?

ayhon:
Yeah, I think it's fine if we remove texan tripes for now

markusdemedeiros:
@Kaptch Sure thing

ayhon:
> One question, what did you mean by "so one doesn't have to write parenthesis for a complex expression"? From my understanding, for expressions we probably always will have some sort of wrapper, similar to hl(..) for HeapLang.

In #397 I had to write `WP (K e) {{ Q }}` for some `fill` lemmas because of the `term:max` priority issue. I hadn't considered that we would probably need some kind of parenthesis for the program expressions regardless. It might not be as blocking as I first thought then (although I would personally still want to know why Lean is failing here regardless)

EDIT: Some light on the syntax issues https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Discussion.3A.20The.20Meta.20Caf.C3.A9/near/596247825

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -0,0 +1,769 @@
...
+variable {hlc : outParam Bool}
+variable {Expr State Obs Val : Type _}
+variable [Λ : Language Expr State Obs Val]
+variable {GF : BundledGFunctors}
+variable [ι : IrisGS_gen hlc Expr GF]
+
+variable {s : Stuckness} {E : CoPset} {e : Expr} {v : Val} {Φ Ψ : Val → IProp GF} {P Q R : IProp GF}
+
+instance frameWp {p : Bool} [H : ∀ v, Frame p R (Φ v) (Ψ v)] :
+    -- TODO: I didn't move over the `FrameInstantiateExistDisabled` constant. Ask if it's necessary.
```
MackieLoeffel: `FrameInstantiateExistDisabled` does not exist at the moment. You can leave it off here, but maybe add a TODO comment to add it once https://github.com/leanprover-community/iris-lean/issues/390 is implemented

Iris/Iris/ProofMode/ProofModeM.lean
```diff
@@ -70,6 +70,17 @@ def addBIGoalWithoutFVars {prop : Q(Type u)} {bi : Q(BI $prop)}
   {e} (hyps : Hyps bi e) (goal : Q($prop)) (toClear : Array FVarId) (name : Name := .anonymous) : ProofModeM Q($e ⊢ $goal) := do
   withoutFVars (u:=0) toClear (addBIGoal hyps goal name)
 
+def runTacticWithoutFVars {prop : Q(Type u)} {bi : Q(BI $prop)}
```
MackieLoeffel: This seems to be a left over from some merge. It should be removed.

Iris/Iris/BI/WeakestPre.lean
```diff
@@ -0,0 +1,140 @@
...
+
+syntax wpExpr :=
+  term:max (" @ " term:max (" ; " term:max) <|> ((" ? ")? )) <|> (" ? ")?
+
+declare_syntax_cat wpPostcondInner
+syntax ident ", " term : wpPostcondInner
+syntax term : wpPostcondInner
+
+declare_syntax_cat wpPostcond
+syntax " {{ " wpPostcondInner " }} " : wpPostcond
```
MackieLoeffel: Does this turn "}}" into a token? Can one still write an implicit parameter of some type `T` with a universe parameter  as `{T.{u}}` or does one now have to write it as `{T.{u} }`? (I've run into this issue with a similar notation, but I am not sure how to best fix it there.)

Kaptch:
> > One question, what did you mean by "so one doesn't have to write parenthesis for a complex expression"? From my understanding, for expressions we probably always will have some sort of wrapper, similar to hl(..) for HeapLang.
> 
> In #397 I had to write `WP (K e) {{ Q }}` for some `fill` lemmas because of the `term:max` priority issue. I hadn't considered that we would probably need some kind of parenthesis for the program expressions regardless. It might not be as blocking as I first thought then (although I would personally still want to know why Lean is failing here regardless)
> 
> EDIT: Some light on the syntax issues https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Discussion.3A.20The.20Meta.20Caf.C3.A9/near/596247825

Got it! I'll test tomorrow. 
I misunderstood it initially, and tried to remove parenthesis around compound expressions in hl(..) (before it was required to write (hl(..))). 
Related to that, probably we can now remove parenthesis with from various unexpanders for HeapLang, since it can be handled by parenthesizer and syntax precedences.

Iris/Iris/BI/WeakestPre.lean
ayhon: Can you pass implicit parameters to types using `{T.{u}}`? I feel like I've only ever found that syntax in function literals.

It seems like `}}` is indeed turned into a "token". Running the following
```lean4
run_meta let s ← `(WP 0 {{ a }}); logInfo s!"{repr s}"
```
seems to show `}}`` as represented by
```
Lean.Syntax.atom (Lean.SourceInfo.synthetic { byteIdx := 1944 } { byteIdx := 1952 } false) "}}"
```
However, all "string" components of a `syntax` declaration are translated into `Lean.Syntax.atom`, so this is probably not what you were asking. 

Could you think of an example where this could be an issue? Maybe using `atomic("}}")` would help in that instance

Kaptch: That's a good find!
```example {a : PUnit.{i}} : PUnit.{i} := a```
Does not work without space in between braces. The same issue is with TC argument with universe levels.

Comment (Kaptch) [reply Iris/Iris/BI/WeakestPre.lean]:
```
syntax " {" "{ " wpPostcondInner " }" "} " : wpPostcond
syntax " [" "{ " wpPostcondInner " }" "] " : wpPostcond
``` 
Should work

File: Iris/Iris/BI/WeakestPre.lean [hunk 4]
```diff
@@ -57,8 +57,8 @@ syntax ident ", " term : wpPostcondInner
 syntax term : wpPostcondInner
 
 declare_syntax_cat wpPostcond
-syntax " {{ " wpPostcondInner " }} " : wpPostcond
-syntax " [{ " wpPostcondInner " }] " : wpPostcond
+syntax " {" "{ " wpPostcondInner " }" "} " : wpPostcond
+syntax " [" "{ " wpPostcondInner " }" "] " : wpPostcond
```
ayhon, line 67: Why is this preferable to what we had before?

Iris/Iris/BI/WeakestPre.lean
ayhon: Maybe we could add a comment explaining the rationale here

Iris/Iris/BI/WeakestPre.lean
Kaptch: Instead of having "}}" and "}]" registered as tokens, tokens "}" and "]" are lexed, which avoids conflicts as in the small example mentioned above (without this change, the example gives an error). It becomes then the parser's job to figure out if it's part of variable declaration or WP postcondition

Kaptch:
Hi again! Thank you for a really good PR! 

I have only one minor note, when doing IPM proofs, it is almost always better to use exclusively IPM tactics, unless something is not supported yet. Other than that, all is really good, and it is a really nice contribution! We are getting to the full version very fast! :)

About the (K (v : Val)) in wp_bind. I think it might be possible to make BI.WeakestPre depend of ProgramLogic.Language and have something like (fun ($v : Language.Val) => ...) in the macro for postconditions. I'm not sure if we should commit to that at this point. IMHO, it might be better to get HeapLang ready, and see if it makes sense to get back to it or not. 

@MackieLoeffel I've noticed that icases doesn't support \<_, H\> and \<H, _\> for A \and B hypothesis. I've left a comment in the code (wp_step_fupdN_strong lemma). Is there a better place to track it?

@markusdemedeiros I release the lock on the file :)

markusdemedeiros:
Right on. My turn.

MackieLoeffel:
> @MackieLoeffel I've noticed that icases doesn't support <_, H> and <H, _> for A \and B hypothesis. I've left a comment in the code (wp_step_fupdN_strong lemma). Is there a better place to track it?

The patterns are slightly different in Lean. `_` in Lean corresponds to `?` in Rocq. To get `_` in Rocq, you should use `-` in Lean. Can you try if `<-, H>` works?

markusdemedeiros:
A general comment: please do not state theorems as functions unless you have a good reason to. It's better to give names to all hypotheses so it is easier to cleanup later (implicits, named arguments, etc.)

markusdemedeiros:
Also: no universe names unless you have a really good reason. `Type _` avoids the issue where you reuse a letter and put in a hard to track down universe constraint.

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -0,0 +1,788 @@
...
+    iprop(stateInterp σ ns obs nt ⊢ |={∅}=> stateInterp σ (ns + 1) obs nt)
+
+
+variable {hlc : outParam Bool}
+variable {Expr State Obs Val}
+variable [Λ : Language Expr State Obs Val]
+variable {GF : BundledGFunctors}
+variable [ι : IrisGS_gen hlc Expr GF]
+
+instance : IrisGS_gen hlc Expr GF → Language Expr State Obs Val := fun _ => Λ
```
markusdemedeiros: This is a weird instance, why is it here? Deleting this seems to do nothing atm.

ayhon: Oops, might have been left over from earlier designs. I swear at some point it did do something, but all the better if it's no longer needed.

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -0,0 +1,788 @@
...
+theorem wp_value_fupd' {s : Stuckness} {E} {Φ : Val → IProp GF} {v : Val} :
+    WP (v : Expr) @ s ; E {{ Φ }} ⊣⊢ |={E}=> Φ v := by
+  simp [IProp.ext wp_unfold, toVal_coe, BI.BIBase.BiEntails.rfl, wp.pre]
+
+@[rocq_alias wp_strong_mono]
+theorem wp_strong_mono {s₁ s₂ : Stuckness} {E₁ E₂} {e : Expr} {Φ Ψ : Val → IProp GF} :
+    s₁ ≤ s₂ → E₁ ⊆ E₂ →
+    ⊢ WP e @ s₁ ; E₁ {{ Φ }} -∗ (∀ v, Φ v ={E₂}=∗ Ψ v) -∗ WP e @ s₂ ; E₂ {{ Ψ }} := by
+  intro hs hE
+  iloeb as IH generalizing %e %Φ %Ψ %E₁ %E₂ %hE
```
markusdemedeiros, line 195: Noting a strange bug here: the delaborator seems to fail and we get a bunch of `BI.forall` statements

markusdemedeiros (approved):

markusdemedeiros:
Awesome, this will be great to have in. I looked over the BI code and it looks good enough to me.

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -0,0 +1,689 @@
...
+/-- Carrier typeclass for the `stateInterp` operation.
+
+This operation cannot be placed directly inside of `IrisGS_gen`
+because Lean then wouldn't be able to derive from its arguments
+the values of `Expr` and `Val`, and so they're asked explicitly.
+This was not a problem in Iris Rocq becuse of canonical structures.
+In Iris Lean, we instead fix our choice of `State` from the choice
+of `Expr`, so `Expr` cannot be inferred from `State` instead.
+-/
+class StateInterp (State : semiOutParam $ Type _) (Obs : outParam $ Type _) (GF : BundledGFunctors)
```
MackieLoeffel: Should `State` be a `semiOutParam` or an input parameter? I don't fully understand the comment above, but it seems like `Expr` should infer `State` and thus we can make it an input parameter here? (At the moment it is a bit weird that `GF` is the only input parameter, because there might multiple state interpretations in a single `GF`.

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -0,0 +1,689 @@
...
+      refine OFE.Contractive.distLater_dist fun m m_n => ?_
+      refine BIFUpdate.ne.ne ?_
+      refine step_fupdN_ne.ne ?_
+      refine BIFUpdate.ne.ne ?_
+      refine BI.sep_ne.ne .rfl ?_
+      refine BI.sep_ne.ne ?_ ?_
+      · exact Hwp m m_n _ _ _
+      · exact BI.BigSepL.bigSepL_dist <| fun _ => Hwp m m_n _ _ _
+
+@[implicit_reducible, rocq_alias wp_def]
```
MackieLoeffel: Out of curiosity: Why is the `implicit_reducible` reducible necessary here?

Iris/Iris/Std/GenSets.lean
```diff
@@ -353,10 +356,17 @@ theorem insert_subset_subset {s₁ s₂ : S} {x : A} (H : s₁ ⊆ s₂) : inser
...
   intro x _; assumption
 
+instance : Std.Refl (fun x y : S => x ⊆ y) where
+  refl _ := subset_refl
+
 /-- Subset relation is transitive. -/
 theorem subset_trans {s₁ s₂ s₃ : S} : s₁ ⊆ s₂ → s₂ ⊆ s₃ → s₁ ⊆ s₃ := by
   intro h1 h2 x hx; exact h2 _ (h1 _ hx)
 
+--                                        ↓ It looks like a face 🥹
```
MackieLoeffel, line 366: Best comment in Iris Lean 😄

MackieLoeffel (commented):
This PR looks really great! I had some minor thoughts when looking over it (but none of this should be blocking merging this. (I hope that I will have time to look into the printing problem with iloeb at some point, but not sure when that will exactly be.)

Kaptch:
> This PR looks really great! I had some minor thoughts when looking over it (but none of this should be blocking merging this. (I hope that I will have time to look into the printing problem with iloeb at some point, but not sure when that will exactly be.)

Hi Michael, thanks for the tip with icases \<-, H\>, that's exactly what I wanted!

Please take a look at https://github.com/leanprover-community/iris-lean/pull/393/changes#diff-6e68887200b797a6df93c0284434bf155c23f7c61bea9d8f8452950382bfb2fcR72 
I think the problem was that irevert was keeping originally implicit binders as implicit, which breaks delaboration for forall. I forced it to be explicit here, but I'm not sure if it's a good solution (alternatively, delab for forall can be extended, but having implicit binders for IPM forall seems a bit odd to me). If you think it's not a good solution, I will revert, and the PR can be merged as it is.

Iris/Iris/ProgramLogic/WeakestPre.lean
ayhon: If I understand correctly, this makes `State` be an input parameter as well, it just requires that the typeclass is instantiated with concrete values for `State`. I believe at some point I managed to convince myself that this was an acceptable design decision to make, but since I failed to document why, and dropping the `semiOutParam` doesn't seem to break anything (neither here nor in the lifting PR #397), I think it's best if we leave it out until it becomes apparent why it's needed again (which I'm not longer sure it is)

Iris/Iris/Std/GenSets.lean
ayhon: 😅

Iris/Iris/ProgramLogic/WeakestPre.lean
ayhon: It isn't, I believe it's added automatically by the `instance` keyword (and so it says in the `implicit_reducible` docstring). I believe what might have happened is
 1. I translated [`wp_def`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/program_logic/weakestpre.v#L93) as a regular definition, since in Iris Rocq it's defined as such, not an instance
 2. Lean gave me the error ["Definition `Iris.wp.def` of class type must be marked with `@[reducible]` or `@[implicit_reducible]`"](https://live.lean-lang.org/#codez=MYGwhgzhAEBiD29oHcAWBTATugUNaAZotAFzQByYALtDjgPQC0jOyYmAdgJYcDmZAEXQEeXKl3gdoAAwDqiAHRF406PALRQkGFQCeAB3TQAtgFcINAEZHj7ANboAJijGoZAAQDa2R6eBdLEHQAXVV4TA9PLmN9EC5%2FKgB9Hz8AoNCcRnocAGJeU3ZHROMIXhgeHEdhaHl4JWIyBCQ0LCNlUgBeaAAGIA), so I added `implicit_reducible`
 3. When I went to translate [`wp'`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/program_logic/weakestpre.v#L96-98), which is the actual instance given by Rocq, I decided that actually I wanted `wp.def` to be my instance (since we're not doing the seal / unseal thing happening in Rocq)
 4. I changed `def` to `instance`, making the `@[implicit_reducible]` redundant.

MackieLoeffel:
I've looked into this and your fix seems like the best option. I added some comments and a test for this behavior.

Iris/Iris/NotationTest.lean.ignore
```diff
@@ -0,0 +1,118 @@
+import Lean
+-- import Lean.NotationExtra
```
MackieLoeffel: Should this file be here?

PR #398, opened 2026-05-19
feat: more flexible imod syntax

MackieLoeffel:
## Description
This PR allows a bit more syntax for `imod` without `with`.


markusdemedeiros:
Nice, I had this thought as well but never committed it so +1 from me :)

PR #403, opened 2026-05-22
fix: Rename TotalWP to TotalWp

MackieLoeffel:
~This reduces the need for type annotations and there should be at most one Wp instance per Expr.~

@ayhon Please have a look if this makes sense.


Iris/Iris/BI/WeakestPre.lean
```diff
@@ -45,10 +45,10 @@ instance : Std.IsPreorder Stuckness where
 
 end Stuckness
 
-class Wp (PROP Expr : Type _) (Val : outParam (Type _)) (A : Type _) where
+class Wp (PROP Expr : Type _) (Val : outParam (Type _)) (A : outParam (Type _)) where
   wp : A → CoPset → Expr → (Val → PROP) → PROP
 
-class TotalWP (PROP Expr) (Val : outParam (Type _)) (A : Type _) where
+class TotalWp (PROP Expr) (Val : outParam (Type _)) (A : outParam (Type _)) where
```
ayhon: The reason why there was a `Wp` and `TotalWP` is that, per Iris conventions and Mathlibs naming conventions, both should be spelled out as `WP`. However, this makes the `WP e @ s {{ Q }}` notation clash with the typeclass name, which is why I changed it to `Wp` while I was developing on the file. I think we may be able to fix this by making "WP wpExpr wpPostcond" atomic, so if it cannot be parsed as such it chooses to parse `WP` as an identifier instead. However, I'm not sure what other ramifications making the whole parser atomic could unchain. 

I just wanted to point this out in case we wanted to make the deliberate decision to deviate from normal naming conventions in order to dodge this issue.

ayhon:
I don't really have any experience of using `WP` in as a user to define program logics yet, so if you think making Stuckness an `outParam` makes sense I defer to your expertise.

Iris/Iris/BI/WeakestPre.lean
ayhon: I couldn't find a way to fix it with `atomic` or something of the sort. I guess this is why `Std.Do` defines its notation for `WP` wrapping on the program instead, as `wp⟦$x:term⟧ $Q`

MackieLoeffel: Thanks for pointing this out! It seems like you thought about this quite a bit more than I did. I just thought that `TotalWP` is a typo since `TotalWP` and `Wp` seems inconsistent. (I don't know the mathlib naming convention well enough to say whether `WP` or `Wp` is the right spelling, both seems fine to me.) @markusdemedeiros What are your thoughts about this?

markusdemedeiros: Sorry for missing this! My understanding is that abbreviations are typically capitalized (to be consistent with the fully spelled out name). I don't completely understand the issue, but if we have to deviate from convention for technical reasons that's totally fine.

MackieLoeffel: I guess the question is whether we should deviate from the convention only for `Wp` where it is necessary for technical reasons, or if we should do it for both `Wp` and `TotalWp` for consistency between `Wp` and `TotalWp`.

markusdemedeiros: I think that consistency within Iris makes more sense here (in terms of how I might be be searching anyways). But I think this is quite minor.

MackieLoeffel:
I thought about this again and maybe let's keep A as an input parameter for the moment. There might be cases where this is useful to disambiguate different WPs (though these would be very niche cases). So the only change of this MR is now to change TotalWP to TotalWp for consistency with Wp. @markusdemedeiros If you think this makes sense, you can merge this PR. Otherwise feel free to close it.

MackieLoeffel:
Since Markus seems to agree with this change (https://github.com/leanprover-community/iris-lean/pull/403#discussion_r3313640239), I am merging this.

PR #404, opened 2026-05-23
refactor: use step_fupd_mono more

ayhon:
## Description
Some lemmas in `BI/Updates.lean`, which could use the newly added `step_fupd_mono`.

This lemma was not present in Rcoq, I suspect because the generalized rewriting machinery make it not really needed. However, as of now we don't really have generalized rewriting, so having a `step_fupd_mono` lemma may help when building terms directly.

## Checklist
* [X] I have updated `PORTING.md` as appropriate

### DIFF
--- a/Iris/Iris/BI/Updates.lean
+++ b/Iris/Iris/BI/Updates.lean
@@ -358,6 +358,10 @@ theorem step_fupdN_ne {E1 E2 : CoPset} {n : Nat} :
     | zero => simp [Nat.repeat, xy_i]
     | succ n IH => exact ne.ne (later_ne.ne (ne.ne IH))
 
+theorem step_fupd_mono {Eo Ei : CoPset} {P Q : PROP} :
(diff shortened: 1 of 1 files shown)


markusdemedeiros:
Ping me when this is ready to merge!

ayhon:
@markusdemedeiros I think it's ready to merge now, not really a big PR, just using it in a couple of places in the `step_fupd_*` lemmas.

While looking around to see if I could beef up a bit this PR though I noticed that some of the delaborators for `BI/Updates.lean` definitions were missing. I gave a try at adding them in #408, so there's something more interesting to review than just shuffling definitions. I think they came out fine, but I was forced to use an `[app_delab` to properly handle the delaboration of the `Nat.iterate` lambda. There might be a simpler way of working around this issue, but I couldn't find it

markusdemedeiros:
Little refactoring PR's like this one are very very welcome btw!

PR #407, opened 2026-05-24
Start porting GhostMap

### LINKED ISSUE #271: Port base_logic/lib/ghost_map.v
## Summary

Port [`base_logic/lib/ghost_map.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/lib/ghost_map.v).

## Dependencies

**Rocq dependencies:**

- `proofmode/proofmode` (aggregation)
- #252
- #262
- #265
- #223

suhr:
## Description

Start porting ghost_map (slowly).

Fixes #271

## Checklist


Kaptch (commented):
You can improve tc inference if you make H an outParam: 
```
class GhostMapG (GF : BundledGFunctors)(F K V: Type _)(H : outParam <| Type _ → Type _)
    [UFraction F][LawfulPartialMap H K] where
  elem: ElemG GF (constOF (HeapView F K (Agree (LeibnizO V)) H))
```

Then, a lot of notation can be used if you specify GF that has GhostMapG, e.g.:
```
@[rocq_alias ghost_map_elem_timeless]
instance (γ : GName)(k: K)(dq: DFrac F)(v: V): BI.Timeless (PROP := IProp GF) (γ ↪◯MAP[k]{dq} v) :=
  iOwn_timeless (E := hgm.elem)
```

Upd.: that should also allow to state BI proof rule, while retaining notations:
```
\vdash@{IProp GF} <notation here>
```

suhr:
Thanks! I also turned `F` into an `outParam`, and it seems to work fine.

Iris/Iris/Instances/Lib/GhostMap.lean
```diff
@@ -0,0 +1,235 @@
...
+notation γ " ↪◯MAP[" k "] " v => ghost_map_elem γ (DFrac.own 1) k v
+
+section lemmas
+
+variable (F K V : Type _) (H : Type _ → Type _) [UFraction F] [lpm: LawfulPartialMap H K] [CMRA V]
+variable [hgm: GhostMapG GF F K V H]
+
+@[rocq_alias ghost_map_elem_timeless]
+instance (γ : GName)(k: K)(dq: DFrac F)(v: V): BI.Timeless (PROP := IProp GF) (γ ↪◯MAP[k]{dq} v) :=
+  iOwn_timeless (E := hgm.elem)
```
Kaptch: Instead of giving names to instances, and using `hgm.elem`, usually it is prefered to use `GhostMapG.elem`, while keeping instances anonymous
```suggestion
  iOwn_timeless (E := GhostMapG.elem)
```

Iris/Iris/Instances/Lib/GhostMap.lean
```diff
@@ -0,0 +1,235 @@
...
+
+@[rocq_alias ghost_map_elem_as_fractional]
+instance (γ : GName) (k: K) (v: V)
+    : AsFractional (PROP := IProp GF) (γ ↪◯MAP[k]{.own q} v)
+      (fun q => γ ↪◯MAP[k]{.own q} v) q := sorry
+
+@[rocq_alias ghost_map_elems_unseal]
+theorem ghost_map_elems_unseal [LawfulFiniteMap H K] γ (m : H V) dq :
+  ([∗map] k ↦ v ∈ m, γ ↪◯MAP[k]{dq} v) ==∗
+  iOwn (E := hgm.elem) γ ([^ CMRA.op map] k ↦ v ∈ m,
```
Kaptch: Oh, i see. Here for some reason op can't find a camera if GhostMapG.elem is used.

Iris/Iris/Instances/Lib/GhostMap.lean
```diff
@@ -0,0 +1,235 @@
...
+    Frag (V:= Agree (LeibnizO V)) k dq (toAgree { car := v })) := sorry
+
+@[rocq_alias ghost_map_elem_valid]
+theorem ghost_map_elem_valid (γ : GName) (k : K) (dq: DFrac F) (v: V) :
+  ⊢@{IProp GF} (γ ↪◯MAP[k]{dq} v) -∗ ⌜✓ dq⌝ := sorry
```
ayhon:
```suggestion
  ⊢@{IProp GF} (γ ↪◯MAP[k]{dq} v) -∗ internalCmraValid dq := sorry
```
There's two definitions of validity for resources in Iris. One works over `Prop`, which is the one used here. However, we can be more precise by using the one which works for an `Sbi PROP`, the internal validity predicate, accessible through [`internalCmraValid`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/BI/Cmra.lean#L30)

Iris/Iris/Instances/Lib/GhostMap.lean
```diff
@@ -0,0 +1,235 @@
...
+
+@[rocq_alias ghost_map_elem_valid_2]
+theorem ghost_map_elem_valid_2 (γ : GName) (k : K) (dq1: DFrac F) (dq2: DFrac F) (v1: V) (v2: V) :
+  ⊢@{IProp GF} (γ ↪◯MAP[k]{dq1} v1) -∗ (γ ↪◯MAP[k]{dq2} v2) -∗ ⌜✓ (dq1 • dq2) ∧ v1 = v2⌝ := sorry
```
ayhon:
```suggestion
  ⊢@{IProp GF} (γ ↪◯MAP[k]{dq1} v1) -∗ (γ ↪◯MAP[k]{dq2} v2) -∗ internalCmraValid (dq1 • dq2) ∧ ⌜v1 = v2⌝ := sorry
```

Iris/Iris/Instances/Lib/GhostMap.lean
ayhon:
```suggestion
  (γ ↪◯MAP[k]{dq} v) ⊢@{IProp GF} ⌜✓ dq⌝ := sorry
```
Is there a reason to prefer the `EmpValid` formulation of the theorem? Otherwise, I'd say to put the `⊢` in place of the `wand`, since it makes using (i.e.) `BIBase.Entails.trans` more convenient.

Kaptch: @ayhon I think there might be a misunderstanding here. I think the [original file](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/lib/ghost_map.v#L91) does have pure proposition brackets.

ayhon: Ok, I see what happened. I was mixing [the old version](https://gitlab.mpi-sws.org/iris/iris/-/blob/61840d318d6e147f3d8b8bb635b7235ee6d9c501/theories/base_logic/lib/gen_heap.v#L110-119) of the `gen_map.v` file with the new version, which used to be defined with a `✓` over `PROP`, but which is now over `Prop`. 

On a separate note, why is this?

Iris/Iris/Instances/Lib/GhostMap.lean
```diff
@@ -0,0 +1,235 @@
...
+  get? m k = None →
+  ⊢@{IProp GF} (γ ↪●MAP m) ==∗ (γ ↪●MAP insert m k v) ∗ (γ ↪◯MAP[k]{.discard} v) := sorry
+
+@[rocq_alias ghost_map_delete]
+theorem ghost_map_delete {γ} {m: H V} (k: K) (v: V) :
+  ⊢@{IProp GF} (γ ↪●MAP m) -∗ (γ ↪◯MAP[k] v) ==∗ γ ↪●MAP delete m k := sorry
+
+@[rocq_alias ghost_map_update]
+theorem ghost_map_update {γ} {m: H V} (k: K) (v: V) (w: V) :
+  ⊢@{IProp GF} (γ ↪●MAP m) -∗ (γ ↪◯MAP[k] v) ==∗ (γ ↪●MAP insert m k v) ∗ γ ↪◯MAP[k] w := sorry
```
ayhon:
```suggestion
@[rocq_alias ghost_map_update]
theorem ghost_map_update {γ} {m: H V} (k: K) (v: V) (w: V) :
  ⊢@{IProp GF} (γ ↪●MAP m) -∗ (γ ↪◯MAP[k] v) ==∗ (γ ↪●MAP insert m k w) ∗ γ ↪◯MAP[k] w := sorry
```

Iris/Iris/Instances/Lib/GhostMap.lean
Kaptch: I don't know, it was before I learned about Iris. Ralf might remember, the snapshot that I've sent was edited by him.

ayhon: Fair enough, I'll ask him tomorrow

Iris/Iris/Instances/Lib/GhostMap.lean
suhr: So using `hgm.elem` everywhere might be a more consistent option.

Kaptch:
Hi again! 
Thanks for working on this issue! 
I made a pass (I will also duplicate this message on Zulip, in case someone later has similar questions).
First, and update to @ayhon: I made one change to the API, since Markus said that the usecase for maps with infinite domains need only HeapView, LawfulPartialMap is changed to LawfulFiniteMap. I hope it's not too breaking for your PR.

There are some comments in the code, where I changed something, but the summary is below:
- Most problems with implicit arguments disappear, if one stays in the Proof Mode as long as possible. That is, `have H := BI.Entails.trans G J` would require a lot of arguments for `G` and `J`, because the proof assistant doesn't know what `PROP` is used, what camera is used, and so on. If something is not supported by the Proof Mode, it is still preferable to use backward reasoning and rely on `refine`/`apply` tactics, so that proof assistant can fill the holes (e.g., `ghost_map_elem_fractional` and `ghost_map_elem_combine`). In particular, I understand why proving `ghost_map_elem_combine` is frustrating, since it relies a lot on rewriting. In this case, applying `iOwn_mono` and constructing a proof of `Included` and `\==` can help.  
- The Proof Mode can help a lot if one adopts intro patterns. In particular, when results of `icases` are pure/persistent/plain `#` allows to retain assumptions, while also getting information out of `icases` invocation (e.g., `ghost_map_elem_frac_ne` and `ghost_map_elem_agree`). In general, the patterns mostly follow Rocq style, but in case of errors, feel free to ask. It would help us a lot when writing documentation as well :)  
- While looking at it, I've noticed that we don't have lemmas related to interactions of `iOwn` and `bigOp*`. While it's possible to use `bigOpM_weak_hom` instead, probably it's better to wait for big op lemmas. Zongyuan kindly agreed to help with porting them.

lzy0505:
Thanks for the PR. I made a pass over it and changed many things. Please have a look at my commit to see if they all make sense.

Here are the items that I didn't fully address in my pass: 
1. Golf the proofs of *_big lemmas.
2. Many partial map side goals are discharged by first unfolding all the way to the basics, then working with them. I think they should be handled by either applying map lemmas, as in Rocq, or using Lean automation. For the latter, it would be nice to add some local simp/grind attributes.

I don't think these are critical, and because we are doing the HeapLang sprint, this PR can be merged anytime!

lzy0505 (approved):

markusdemedeiros:
If it's all right with you folks I'd like to take a quick look over it too. I'll be quick.

markusdemedeiros:
OK, looks good to me. I didn't touch the changes outside of GhostMap, but I was able to cut/abstract away like 100 lines.

Kaptch:
Thank you everyone!

PR #424, opened 2026-05-29
feat: add to_eq to simplify use of eq_of_eqv

MackieLoeffel:
This PR allows to use `x.to_eq` instead of `eq_of_eqv x` and `eq_of_eqv <| equiv_iff.mpr x`, cleaning up some code and in general making it nicer to use Leibniz equality.


lzy0505:
We also have `IProp.ext`. Rename it?

MackieLoeffel:
Good point, I also replaced the usages of `IProp.ext` with `to_eq`.

markusdemedeiros:
Nice!

PR #425, opened 2026-05-30
Port ReservationMap

### LINKED ISSUE #258: Port algebra/reservation_map.v
## Summary

Port [`algebra/reservation_map.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/reservation_map.v).

## Subtasks

- [ ] CMRA
- [ ] Updates

## Dependencies

**Rocq dependencies:**

- #243
- #254
- #229
- #231

suhr:
## Description

Fixes #258 

## Checklist


Iris/Iris/Algebra/ReservationMap.lean
```diff
@@ -0,0 +1,564 @@
...
+  | inl h =>
+    simp only [data, token, op_dataProj, Heap.get?_op, get?_empty] at h
+    exact .inl <| Option.eq_none_of_op_eq_none_left h
+  | inr h' =>
+    simp only [data, token, op_tokenProj] at h'
+    rw [mem_iff_of_valid_union, not_or] at h'
+    · exact .inr <| fun hh => h'.right hh
+    · exact valid_of_eqv (pcore_op_left' rfl).symm valid_set
+
+theorem disj_of_valid_data_op_token (a : H A) (b : CoPset)
```
Kaptch: Is this theorem used anywhere? I couldn't find a similar one in the Rocq version.

suhr: A similar one, `disj_of_validN_data_op_token`, is used in proofs of `alloc` and `updateP`. These lemmas are indeed not a part of Rocq version as well as the definition of `ReservationMap.data`. I just noticed some algebraic properties that look useful and decided to generalize.

Kaptch: Okay, thanks for clarifying!

Kaptch:
Thanks for the PR! All good from my side! 

Relatively high-level comment, but in general, please try to minimize number of statements posed with `have`.
The reason is that it makes it easier to trace state in individual proofs, and minimizes required annotations. 
If they help to simplify proofs (e.g., by supplying them to `grind`), then it might be better to move them closer to the place, where they are used, so it's easier to find what is needed and where.

@markusdemedeiros I'm done with my pass. A bit unrelated to the PR, I've noticed that some DiscreteE instances are marked as theorems, as some -- as instances. Is there some convention about that?

Kaptch (approved):

suhr:
I'm a big fan of spamming `by apply?` and this tends to result in a lot of `have`s. I also find that they make some proofs easier to read to me.

I understand that Iris-Lean proof style is different but I'm not very accustomed to that style.

markusdemedeiros:
I do the same! However these proofs can be more difficult to maintain so we try to inline them afterwards where it makes sense.

markusdemedeiros:
> I'm done with my pass. A bit unrelated to the PR, I've noticed that some DiscreteE instances are marked as theorems, as some -- as instances. Is there some convention about that?

DiscreteE: not to my knowledge, I'd assume that these should all be intances. Check the git blame to see who touched it last though, they might have a reason. Some nonexpansive theorems on the other hand are intentionally not instances.

markusdemedeiros (approved):

markusdemedeiros:
Looks great! Aside from style fixes, the only major improvement was combining your `of_parts` and `unfold` theorems into `iff`. 

Please look over my changes when you have time. There is more golfing to do in the updates section but I will merge as it is in the hot path of several projects.

PR #430, opened 2026-05-31
feat: iinduction

### LINKED ISSUE #360: Port iInduction

alvinylt:
## Description

Implements the `iinduction` tactic.

Addresses #360.

This tactic uses the type class `IntoIH` and its instances, which are ported from [`coq_tactics.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/coq_tactics.v) in the Rocq version. ~~Two~~ *One* of them (~~`intoIH_listForall` and~~ `intoIH_listForall₂`) ~~are~~ *is* newly created in this PR. The type class itself and other instances were already created by @HaleOIC, and have now been moved from [Induction.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Tactics/Induction.lean) to [Classes.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Classes.lean) and [Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean), respectively.

| Type class/type class instance in Lean | Located in                                                                                                                                           | Corresponding definition in Rocq |
|----------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------|
| `IntoIH`                               | [Classes.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Classes.lean)     | `IntoIH`                         |
| `intoIH_entails`                       | [Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean) | `into_ih_entails`                |
| `intoIH_forall`                        | [Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean) | `into_ih_forall`                 |
| `intoIH_imp`                           | [Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean) | `into_ih_impl`                   |
| ~~`intoIH_listForall`~~                    | ~~[Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean)~~ | ~~`into_ih_Forall`~~                 |
| `intoIH_listForall₂`                   | [Instances.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Instances.lean) | `into_ih_Forall2`                |

The mechanism is basically the same as those in Rocq, but the tactic syntax for induction is quite different in Lean. The implementation of `iinduction` therefore intends to follow [the syntax of `induction` as detailed in the language specifications of Lean](https://lean-lang.org/doc/reference/latest/Tactic-Proofs/Tactic-Reference/#induction):

- [x] `iinduction e`
- [x] `iinduction e using r`
- [x] `iinduction e generalizing z₁ ... zₙ`: note that `z₁ ... zₙ` is implemented as selection patterns (`SelPat`) and, as a result, there are no restrictions regarding the order of regular Lean variables and Iris hypothesis names.
- [x] `iinduction e with | constr₁ => tac₁ | ... | constrₙ => tacₙ`: the constructor names can either be long (e.g., `Nat.zero`) or short (e.g., `zero`). Arguments are optionally given names or otherwise remain inaccessible.

In a nutshell, the `iinduction` tactic performs the following steps:
1. Generalise the expression `e` if it does not directly correspond to an `FVarId` value.
1. *(Optional, when `generalizing` is used)* Revert Iris hypotheses explicitly specified by the user.
2. Reverts all hypotheses in the spatial context as well as hypotheses in the intuitionistic context involving the variable on which induction is being applied.
    - Hypotheses already reverted in step 1 do not need to be handled in this step.
3. *(Optional, when `generalizing` is used)* Revert pure Lean variables explicitly specified by the user.
4. *(Optional, when `with` is used)* Check that the constructor patterns (alternative names) provided by the user are valid.
5. Uses `Lean.Meta.Tactic.induction` to obtain the induction subgoals.
6. Moves the newly generated induction hypotheses from the regular Lean context into the intuitionistic context.
7. Introduces the reverted hypotheses from steps 1, 2 and 3 back into the spatial and intuitionistic contexts.
8. Create the induction subgoals for the tactic user.
9. Fill in the metavariable for each induction subgoal generated by Lean using the corresponding proof obtained in step 8.
10. *(Optional, when `with` is used)* Handle the tactic sequences for each induction subgoal.
11. Fill in the metavariable for the original proof goal before applying the tactic.

The tactic syntax can be used in combination:
- [x] `iinduction e using r with | constr₁ => tac₁ | ... | constrₙ => tacₙ`
- [x]  `iinduction e using r generalizing z₁ ... zₙ`
- [x]  `iinduction e generalizing z₁ ... zₙ with | constr₁ => tac₁ | ... | constrₙ => tacₙ`
- [x]  `iinduction e using r generalizing z₁ ... zₙ with | constr₁ => tac₁ | ... | constrₙ => tacₙ`

Similar to the built-in `induction` tactic in Lean, the variables and induction hypotheses are assigned inaccessible names unless the user specifies them using the `with` syntax. For example, by using `iinduction n` where `n : Nat`, one may get `n✝` and `n_ih✝` in the step case. Meanwhile, the following assigns the variable and the induction hypothesis the names `n` and `ih`, respectively.

```lean4
iinduction n with
| zero      => ...
| succ n ih => ...
```

One can also use underscores for the variables to remain inaccessible. This is made possible with Lean's `binderIdent`. For example, one can get the unnamed variable `n✝` and the named induction hypothesis `ih`:
```
iinduction n with
| zero      => ...
| succ _ ih => ...
```

Several functions in [Induction.lean](https://github.com/leanprover-community/iris-lean/blob/e376e7047db5551add8b0646ed4646fe07aa22fc/Iris/Iris/ProofMode/Tactics/Induction.lean) are introduced to parse the syntax and handle the constructor names properly. This handles subtle caveats such as duplicate, missing and/or invalid constructor names.

## Checklist


Iris/Iris/ProofMode/Instances.lean
```diff
@@ -917,3 +918,69 @@ instance combineSepGives_persistently [BI PROP] (Q1 Q2 P : PROP)
...
+  into_ih := by
+    intro hImp
+    apply wand_intro
+    refine (sep_mono_r h1.make_affinely.mpr).trans ?_
+    refine persistent_and_affinely_sep_r.2.trans ?_
+    exact pure_elim_r (fun hφ => h2.into_ih (hImp hφ))
+
+/-- Support for induction principles whose IH is guarded by `List.Forall`, e.g.
+    `∀ l, Forall P l → P (Tree l)` arising from nested inductive types like
+    `inductive ntree := Tree : List ntree → ntree`. -/
```
MackieLoeffel: Please add a test for doing an induction on a type like `ntree`.

Iris/Iris/ProofMode/Instances.lean
```diff
@@ -917,3 +918,69 @@ instance combineSepGives_persistently [BI PROP] (Q1 Q2 P : PROP)
...
+      apply intuitionistically_sep_idem.mpr.trans
+      refine sep_mono ?_ ?_
+      · exact intuitionistically_intro' ((h x).into_ih hx)
+      · apply ih
+        apply (List.all_eq_true.mpr)
+        apply hxs
+
+/-- Support for induction principles whose IH is guarded by `List.Forall₂`, e.g.
+    arising from mutual inductive types relating two lists element-wise. -/
+@[rocq_alias into_ih_Forall2]
```
MackieLoeffel, line 954: What Lean type will generate an induction principle that uses List.Forall₂? Please add a test case that tests this lemma.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,506 @@
...
+  ProofModeM.runTactic λ mvar { hyps, goal, .. } => do
+    -- Parse the user-supplied list of variables to be generalised
+    let genSelPats ← liftMacroM <| SelPat.parse genSelPats
+    let genSelTargets ← SelPat.resolve hyps genSelPats
+
+    let pf ← iInductionCore hyps goal fvar none recName genSelTargets
+    mvar.assign pf
+
+elab "iinduction" colGt x:term "using" r:ident
+    "generalizing" genSelPats:(colGt selPat)+ "with" alts:(colGe inductionAlts)* : tactic => do
```
MackieLoeffel: There are a lot of variants of `iinduction`. Can you combine some of these cases using either optional syntax or macros to desugar one to the other? (E.g. can one desugar the cases without "generalizing" to the cases with an empty generalizing clause?)

MackieLoeffel (commented):
I had a brief look and added some comments. I hope to be able to do an in-depth review later.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
-public meta section
-open Lean Elab Tactic Meta Qq
+  Meanwhile, the actual induction subgoal generated by Lean with the `induction`
+  tactic has induction hypotheses in the regular context.
+
+  Given all hypotheses in the Iris proof state (represented by `P`) exist in
+  the intuitionistic context (that is, the spatial context is empty), we can
+  introduce another induction hypothesis generated by Lean into the
+  intuitionistic context.
+-/
```
MackieLoeffel, line 37: What does this comment refer to? It is attached to `revert_IH`, but I don't see the connection between the comment and the statement of `revert_IH`.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+
+/--
+  Given a collection of hypotheses (`hyps`) and a free variable `fvar`, return
+  the subset of intuitionistic hypotheses in `hyps` that contains the `fvar`
+  and all spatial hypotheses in the proof state.
+
+  This function is used for identifying the hypotheses in the Iris proof state
+  that must be reverted before applying Lean's built-in `induction` tactic.
+-/
+private def iHypsContaining {u} {prop : Q(Type $u)} {bi} {e : Q($prop)}
```
MackieLoeffel: The docstring is wrong. This function does not just return a `subset of intuitionistic hypotheses`. It also returns all spatial hypotheses. Maybe rename this function to something like `hypsToGeneralize`?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+  Search for hypotheses in the regular Lean context that represent induction
+  hypotheses and return their IDs as a list.
+-/
+private def findIHs (m : MVarId) : ProofModeM (List FVarId) :=
+  m.withContext do
+    let lctx ← getLCtx
+    let mut ihs := []
+    for decl in lctx do
+      let type ← instantiateMVars decl.type
+      if isIrisGoalWithForalls type then
```
MackieLoeffel: Instead of introducing `isIrisGoalWithForalls`, you can just use `isIrisGoal type.getForallBody`

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+    match hyps.getDecl? ivar with
+    | some ⟨_, _, _, ty⟩ => fvars.any ty.containsFVar
+    | none => false
+
+  (ivars ++ hyps.spatialIVarIds).map ({ kind := .ipm ·, explicit := false })
+
+/--
+  Search for hypotheses in the regular Lean context that represent induction
+  hypotheses and return their IDs as a list.
+-/
```
MackieLoeffel: The key information is missing from this comment: How do we identify what is an induction hypothesis? The answer is that it is an hypothesis that is an Iris Goal. We know that these are the induction hypotheses since otherwise there should never be assumptions that are Iris Goals.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+    (pfIntHyps : Q($e ⊢ □ $e))
+    (φ : Q(Prop)) (p : Q($φ)) (hFVar : FVarId) :
+    ProofModeM (@InductionState u prop bi e) := do
+  let oldE := st.newE
+  let oldHyps : Hyps bi oldE := st.newHyps
+
+  -- Obtain the proposition to be introduced into the intuitionistic context
+  let Q ← mkFreshExprMVarQ q($prop)
+  let some inst ← ProofModeM.trySynthInstanceQ q(IntoIH $φ $e $Q)
+  | throwError "iinduction: type class synthesis with IntoIH failed"
```
MackieLoeffel: This error should be more specific. In particular, it should say which hypothesis could not be converted.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+      if isIrisGoalWithForalls type then
+        ihs := decl.fvarId :: ihs
+    return ihs.reverse
+
+/--
+  Used for every induction hypothesis generated by Lean upon using the built-in
+  `induction` tactic. This function is called by `addIHs` and returns
+  an instance of `InductionState` after handling one induction hypothesis.
+-/
+private def addIH {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e}
```
MackieLoeffel: I think this function can be inlined into `addIHs`.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+      throwError "iinduction: spatial context is not empty after reverting hypotheses"
+  | .sep _ _ _ _ lhs rhs => do
+    let pfL ← buildPfIntHyps lhs
+    let pfR ← buildPfIntHyps rhs
+    pure q((sep_mono $pfL $pfR).trans intuitionistically_sep_2)
+
+/--
+  Tactic syntax for user-supplied alternative names.
+-/
+syntax inductionAlts := "| " binderIdent+ " => " tacticSeq
```
MackieLoeffel: Can you reuse the syntax used by the induction tactic?

MackieLoeffel: Please move all syntax related things to a separate section at the top of the file.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+
+  return st
+
+/--
+  Given hypothesis `hyps` representing `e` where every hypothesis exist in the
+  intuitionistic context, return the proof of `e ⊢ □ e`. Throw an error if
+  `hyps` contains hypotheses in the spatial context.
+-/
+private def buildPfIntHyps {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e}
+    (hyps : Hyps bi e) : ProofModeM Q($e ⊢ □ $e) :=
```
MackieLoeffel: This function does not need to be in ProofModeM, it can simply return an `Option`. Also please give it a more descriptive name and move it to `Expr.lean`.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+  let spatialAndIntuitionisticTargets := iHypsContaining hyps <| pureFVars.concat fvar
+
+  let spatialAndIntuitionisticTargets :=
+    spatialAndIntuitionisticTargets.filter (not <| (explicitIrisTargets.map (·.kind)).contains ·.kind)
+
+  -- Pure hypotheses to be reverted as specified by the user using the `generalizing` syntax
+  let pureTargets : List SelTarget := pureFVars.map ({ kind := .pure ·, explicit := true })
+
+  -- Those in `explicitIrisTargets` get reverted first, so they are last in the list
+  let targets := pureTargets ++ spatialAndIntuitionisticTargets ++ explicitIrisTargets
```
MackieLoeffel: All of the above is quite a long dance to get the desired SelTargets. What if you change the `intuitionistic` SelPat to have an `(containing : Option (List FVarId))` argument to allow selecting all `intuitionistic` hypotheses that contain a specific variable?
I hope that you should then be able to replace all of the above with `SelPat.resolve (genSelTargets ++ [.intuitionistic (some pureFVars.concat fvar), .spatial]`.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+          let st ← addIHs pfIntHyps irisGoal.hyps ihFVars
+
+          -- Generate the induction subgoal for the user
+          let pf' ← withoutFVars (u := 0) ihFVars.toArray <| k st.newHyps irisGoal.goal
+
+          -- Fill the metavariable for the induction subgoal generated by Lean
+          s.mvarId.assign <| q($(st.pf).trans $pf')
+
+          -- Label the induction subgoal for the user with a name
+          let biGoalMVar := (← getThe ProofModeM.State).goals.back!
```
MackieLoeffel: You should not directly access the state of ProofModeM. Why does the code do this and can it be avoided?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+          biGoalMVar.setTag ctor
+
+          -- Handle user-supplied tactics for specific constructors
+          match parsedAlts with
+          | none => pure ()
+          | some parsedAlts =>
+            modify <| fun s => { s with goals := s.goals.pop }
```
MackieLoeffel: Same comment as above.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,514 @@
...
+      return m
+
+  return pf
+
+/--
+  Given a term expression, check whether it is a `FVarId` value. If so,
+  return it directly. Otherwise, generalise the term expressions before
+  returning its `FVarId`.
+-/
+private def termToFVar (x : TSyntax `term) : TacticM FVarId := do
```
MackieLoeffel: The name of this function should mention that it generalizes the term if necessary. Also it is important to mention in the comment that this function must be called before we enter ProofModeM.

Iris/Iris/ProofMode/Expr.lean
```diff
@@ -189,6 +189,12 @@ partial def Hyps.intuitionisticIVarIds {u prop bi} :
   | _, .hyp _ _ ivar p _ _ => if isTrue p then [ivar] else []
   | _, .sep _ _ _ _ lhs rhs => lhs.intuitionisticIVarIds ++ rhs.intuitionisticIVarIds
 
+partial def Hyps.intuitionisticProps {u prop bi} :
```
MackieLoeffel: It seems like this function is never used.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2776,3 +2715,95 @@ example (P Q : PROP) :
...
+
+  Meanwhile, `HQ` is included as a premise of the induction hypothesis without
+  `Q` being universally quantified.
+
+  Note that the following variants of the tactic all produce equivalent subgoals:
+  - `induction n generalizing %P HP HQ %R`
+  - `induction n generalizing %P HQ %R HR`
+  - `induction n generalizing %P HP HQ %R HR`
+  - the tactics above with any permutations of `generalizing` arguments.
+-/
```
MackieLoeffel, line 3012: I don't think this is the right place to add the comment explaining `iinduction`. Instead, this should either be the doc string of `iinduction` or added to tactics.md or both.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2776,3 +2715,95 @@ example (P Q : PROP) :
...
+  | zero => itrivial
+  | ind n ih => itrivial
+
+inductive Tree (α : Type u) where
+  | leaf : Tree α
+  | node : Tree α → α → Tree α → Tree α
+  deriving Repr
+
+example [BI PROP] {α} {t : Tree α} {P : Tree α → PROP} :
+    □ P .leaf -∗ □ (∀ l x r, P l -∗ P r -∗ P (.node l x r)) -∗ P t := by
```
MackieLoeffel: Please add some more tests for the Tree example, e.g. with pure and spatial hypotheses.

MackieLoeffel (commented):
I looked over the PR and added some comments. I have not yet looked at the syntax related things in detail yet.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved, the comment is updated and the function is renamed.

Commit: [8412da6](https://github.com/leanprover-community/iris-lean/pull/430/commits/8412da68973e847974b4d364f84022c81cae58eb)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [4d51583](https://github.com/leanprover-community/iris-lean/pull/430/commits/4d51583160e80ed1edd4d49acd39258205e70fa4).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: I have rephrased the comment to include the detail.

Commit: [4d51583](https://github.com/leanprover-community/iris-lean/pull/430/commits/4d51583160e80ed1edd4d49acd39258205e70fa4)

Iris/Iris/ProofMode/Expr.lean
alvinylt: This function was originally intended to search for the hypotheses that exist in the intuitionistic context and create the proposition `e`, which would then be used to show that `e ⊢ □ e`. Later the function `Hyps.buildIntuitionisticProof` was created to build the proof directly by iteration on `Hyps`, and this function is no longer necessary. I have now removed it.

Commit: [2abdf66](https://github.com/leanprover-community/iris-lean/pull/430/commits/2abdf66e350b470f3a9d84955e975bc49003a001)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [05b2cb9](https://github.com/leanprover-community/iris-lean/pull/430/commits/05b2cb9db6d1fd06bc800637468999f769986b25). The function is renamed as `Hyps.buildIntuitionisticProof`.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [73e28b6](https://github.com/leanprover-community/iris-lean/pull/430/commits/73e28b66e1dfd1d9ff59f0d146bf43efd0b5b07c).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [5570273](https://github.com/leanprover-community/iris-lean/pull/430/commits/557027314edca2d92e591b4148e69f16cfe70d3e).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: I guess the main issue here was that the comment explained *why* `revert_IH` is used by `addIHs` to update the proof in `InductionState`, but the theorem itself is generic. Also I renamed the variables (`P`, `Q`, `R`, ...) at some point without explaining what they represent when it is used by `addIHs`. I have rewritten the explanation, which hopefully removes some confusion.

Commit: [97f0ef7](https://github.com/leanprover-community/iris-lean/pull/430/commits/97f0ef7b25337b841fdf462551d40db318fffb37)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: The function is renamed, and more details are added in the comment.

Commit: [d7d9da5](https://github.com/leanprover-community/iris-lean/pull/430/commits/d7d9da5cf46ea976d4a3400c2f0b3d5be635f89b)

lzy0505:
Great work! If you want to test the tactic further, there are some proofs in `/BI` and `/Instance` that are currently doing `induction`, which can be replaced with the new tactic.
Also don't forget to update `ProofMode/Porting.lean`.

alvinylt:
> Great work! If you want to test the tactic further, there are some proofs in `/BI` and `/Instance` that are currently doing `induction`, which can be replaced with the new tactic. Also don't forget to update `ProofMode/Porting.lean`.

Sure we can see if existing inductive proofs can be simplified.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: I'm trying to extend `SelPat` and `SelPat.resolveOne` as follows.

```lean4
inductive SelPat
  | pure
  | intuitionistic (fvars : Option <| List FVarId)
  | ...
```

```lean4
def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget)
  | ...
  -- Include all hypotheses in the intuitionistic context
  | .intuitionistic none =>
      return hyps.intuitionisticIVarIds.map (⟨.ipm ·, false⟩)
  -- Include hypotheses in the intuitionistic context involving any of the `FVarId` values in the list `fvars`
  | .intuitionistic (some fvars) =>
      let ivars := hyps.intuitionisticIVarIds.filter <| fun ivar =>
      match hyps.getDecl? ivar with
      | some ⟨_, _, _, ty⟩ => fvars.any ty.containsFVar
      | none => false
      return ivars.map (⟨.ipm ·, false⟩)
  | ...
```

In order to find `pureFVars`, we have to first parse `genSelTargets`. I think this means we will have to call `SelPat.resolve` twice and apply `eraseDupsBy` to replace duplicates. This is what I have got so far:

```lean4
private def iInductionCore {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e}
    ... (genSelPats : Option <| List SelPat) :
    ProofModeM Q($e ⊢ $goal) := do
  let genSelTargets : List SelTarget ← do match genSelPats with
  | none => pure []
  | some genSelPats => SelPat.resolve hyps genSelPats

  -- `FVarID` values of hypotheses specified by the user using the `generalizing` syntax
  let pureFVars : List FVarId := genSelTargets.filterMap <|
    fun t => match t.kind with | .pure f => some f | _ => none

  let targets := (genSelTargets ++ (← SelPat.resolve hyps [.intuitionistic (pureFVars.concat fvar), .spatial])).eraseDupsBy (λ a b => a.kind == b.kind)
```

Moreover, this is under the assumption that the order of targets does not matter. In the Rocq version, Iris hypotheses explicitly specified using the `generalizing` syntax are first reverted, which are then followed by the remaining Iris hypotheses and finally pure hypotheses. My original design in Lean is designed to follow the same order, which is why I split the targets into pure ones and IPM kinds. Does the order still matter in Lean? If so, I reckon the new approach might be more complicated than the original one.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: For this particular line about naming the induction subgoal, yes. I have added the name of the subgoal as a new parameter for the continuation function `k`. In `iRevertIntro`, the name is then passed onto `addBIGoal`.

```lean4
k hyps goal fun hyps goal name => do
  iIntroCore hyps goal names (addBIGoal · · name)
```

Commit: [343af53](https://github.com/leanprover-community/iris-lean/pull/430/commits/343af538c119e9efd13c56365270c51dac5ee45d)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Unlike the point above about giving induction subgoals names, I'm not sure whether this part is avoidable. The sequence of tactics supplied by the user for each induction subgoal is part of the `iinduction` syntax, which is applied *after* introducing the induction subgoal using `addBIGoal`.

The built-in `induction` tactic in Lean uses [`evalAlt` in `Lean.Elab.Tactic.Induction`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Tactic/Induction.lean#L79-L103), which has a similar approach of directly pushing and popping goals in the proof state. We can possibly reuse the syntax and functions for the built-in `induction` tactic to avoid repetitive code, but I'm not sure the manipulation is fundamentally avoidable. This will also require changes to `iRevertIntro` as it will involve steps beyond calling `addBIGoal`. Perhaps we can discuss further tomorrow?

alvinylt: An alternative solution I can think of is to add a new function in [`ProofModeM.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/ProofModeM.lean) so that the pushing and popping are done there rather than in `iInductionCore`.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Yes, the syntax is based on that in [`Init.Tactics`](https://github.com/leanprover/lean4/blob/master/src/Init/Tactics.lean#L956-L965). Depending on how [the issue about manipulating `ProofModeM`](https://github.com/leanprover-community/iris-lean/pull/430#discussion_r3345631943) is solved, perhaps we can reuse functions in the built-in libraries as well.

Iris/Iris/Tests/Tactics.lean
alvinylt: I have moved the explanation of `iinduction` to the elaboration. As a side note, I think it will be great if we add docstring comments to all the elaborations that define the IPM tactics so that information about the tactic is shown by hovering over them.

Tactic documentation shown by hovering over it

Commit: [ff898f6](https://github.com/leanprover-community/iris-lean/pull/430/commits/ff898f65b7e63b855961eb323c17d29a58cf0e8a)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: I have consolidated all the elaboration into one by using `syntax` and `elab_rules`.

Commit: [ff898f6](https://github.com/leanprover-community/iris-lean/pull/430/commits/ff898f65b7e63b855961eb323c17d29a58cf0e8a)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [ff898f6](https://github.com/leanprover-community/iris-lean/pull/430/commits/ff898f65b7e63b855961eb323c17d29a58cf0e8a).

alvinylt:
I have made a couple of changes over the past few days, mainly regarding the syntax of the tactic.

### Reusing `inductionAlts` from the built-in library

The syntax for user-supplied alternative names was originally defined as part of the implementation:

```lean4
syntax inductionAlts := "| " binderIdent+ " => " tacticSeq
```

It turns out that we can reuse the syntax definitions in [`Init.Tactics`](https://github.com/leanprover/lean4/blob/master/src/Init/Tactics.lean#L950-L965). By doing so, the syntax of `iinduction` closer resembles the built-in `induction` tactic. In particular, there are several ways to use the `iinduction e with | constr₁ => tac₁ | ... | constrₙ => tacₙ` syntax that I originally missed:

1.  Using wildcard (`_`) as the last case for all constructors that have not yet been handled. For example, given `n : Nat`, the wildcard is applicable for the step case (`succ`).
```lean4
iinduction n with
| zero => <insert tactics here>
| _ => <insert tactics here>
```
The wildcard must always be the last case and should not be used if all constructors have already been handled. Otherwise, the implementation should be able to throw the corresponding error messages:

```
iinduction: invalid occurrence of the wildcard alternative `| _ => ...`: It must be the last alternative
iinduction: wildcard alternative is not needed
```

2. Handling multiple constructors with the same tactic sequences. Again with an example where `n : Nat`:

```lean4
iinduction n with
| zero | succ n ih => <insert tactics here>
```

3. Using a hole (`_` or `?_`) to defer the case. For example:

```lean4
iinduction n with
| zero => ?_
| succ n ih => <insert step case tactics here>
<insert base case tactics here>
```

4. Applying one tactic after the `with` keyword before splitting into cases. For example:

```lean4
iinduction n with simp
| zero => <insert tactics here>
| succ n ih => <insert tactics here>
```

This one is tricky because this tactic (in the example above, `simp`) may already be sufficient to solve some of the induction subgoals. As a result, some of the cases become redundant, and the implementation is able to throw an error if redundant alternative names are given. Example:

```
iinduction: alternative `zero` is not needed
```

It is also possible that the tactic (e.g. `cases ...`) may generate even more subgoals. From my understanding of how the built-in `induction` tactic works, the tactic user has to provide a sequence of tactics for every constructor, which has to be applicable to all subgoals for that constructor. Now `iinduction` works in the same way.

5. A minor point: one can optionally have `@` before the constructor. Example:

```lean4
iinduction n with simp
| @zero => <insert tactics here>
| @succ n ih => <insert tactics here>
```

This is part of the definition of `inductionAltLHS`, which I think has no effect on tactic behaviour.

### The induction target as `term` instead of `ident`

In an earlier version of the implementation, the induction target is an `ident`. Now it is able to accept any term, which means the following, for example, is now possible:

```lean4
iinduction n + m with ...
```

If the term does not directly correspond to an `FVarId`, it is first generalised.

alvinylt:
Following recent discussions, I have slightly improved the way selection targets are handled.

In particular, the following are automatically reverted before induction on `x` (that is, part of the selected targets for `iRevertIntro`):
- All Iris hypotheses in the spatial context.
- All Iris hypotheses in the intuitionistic context that involve `x`.

The previous implementation also took the user-supplied set of hypotheses for generalisation into consideration. Given `iinduction e generalizing z₁ ... zₙ`, the implementation not only reverted `z₁ ... zₙ` but also all pure/Iris hypotheses that involves any of `z₁ ... zₙ`. As discussed in [one of the comments](https://github.com/leanprover-community/iris-lean/pull/430#discussion_r3345622236), there are quite a few steps to find the filter in pure hypotheses in `z₁ ... zₙ` and to search for the relevant hypotheses that were also automatically generalised.

However, it was problematic because it is possible that a pure/Iris hypothesis `x` depends on `y`, and `y` in turn depends on `z`. The approach above is able to generalise `x` and `y` but not `z`. As a result, the tactic gives an error when `iRevertCore` is used. The main caveat here is that, while the built-in `induction` tactic in Lean is designed to find out all forward-dependent hypotheses and generalise them all, this is supported in neither Rocq version of `iInduction` nor the `irevert` tactic in Lean. As per recent discussions, the latest version of `iinduction` now follows the latter for more predictable tactic behaviour.

Take the following as an example:
```lean4
example [BI PROP] {P : PROP} {m n : Nat} {Q R S : Nat → PROP} :
    ⊢ P -∗ □ Q m -∗ □ R m -∗ □ S n -∗ ⌜n + 0 = n⌝ := by
  iintro HP #HQ #HR #HS
  iinduction n generalizing %m with
```

- `∗HP : P` is automatically reverted because it exists in the spatial context.
- `□HS : S n` is also automatically reverted because it depends on the induction target `n`.
- The user stipulates the generalisation of the pure variable `m`, but `□HQ : Q m` and `□HR : R m` are not included in `generalizing ...`.

The tactic now checks the forward dependencies and offers a fix:

Suggested fix for generalising Iris hypotheses

This also includes checking for pure Lean hypotheses to be generalised. Example:

Suggested fix for generalising pure hypotheses

With this approach, the code to determine the selection targets for reverting is slightly simplified.

alvinylt:
Another point to highlight is that, for the tactic syntax `iinduction e generalizing z₁ ... zₙ`, there is no restriction regarding the order of pure and Iris hypotheses in the selection pattern `z₁ ... zₙ`.

The Rocq version of `iInduction`, pure and Iris hypotheses to be generalised are separate. Iris hypotheses are first generalised and then pure hypotheses comes next.

I had been thinking of restoring the hypotheses in whatever order the user specifies under the `generalizing` clause, which could have further simplified the code for handling selection targets in `iInductionCore`. The problem was that there is tactic could become unpredictable.

For example, let's say we have the following proof:
```
example [BI PROP] {P : Nat → PROP} {Q : PROP} {m n : Nat} :
    ⊢ P m -∗ □ Q -∗ ⌜n + 0 = n⌝ := by
  iintro HP #HQ
  iinduction n generalizing %P HQ HP with
```

The induction hypothesis generated is `∀ P, □ Q -∗ P m -∗ ⌜n + 0 = n⌝`.

We can rearrange the order of hypotheses under the `generalizing` clause:
```
example [BI PROP] {P : Nat → PROP} {Q : PROP} {m n : Nat} :
    ⊢ P m -∗ □ Q -∗ ⌜n + 0 = n⌝ := by
  iintro HP #HQ
  iinduction n generalizing %P HP HQ with
```
Now the induction hypothesis is `∀ P, P m -∗ □ Q -∗ ⌜n + 0 = n⌝`.

Here's a third example:
```
example [BI PROP] {P : Nat → PROP} {Q : PROP} {m n : Nat} :
    ⊢ P m -∗ □ Q -∗ ⌜n + 0 = n⌝ := by
  iintro HP #HQ
  iinduction n generalizing HQ %P HP with
```
The induction hypothesis becomes `□ Q -∗ ∀ P, P m -∗ ⌜n + 0 = n⌝`.

The first and second cases are more or less the same, but the scope of `∀ P` is the different in the third case. I'm not sure whether this is desirable, considering that the tactic also reverts other Iris hypotheses (i.e., those involving the induction target), and the tactic user has no control over the order of those.

I have currently retained the existing approach of first reverting Iris hypotheses and then pure hypotheses in `iInductionCore` with slight improvement to avoid repetitive code:
```lean4
let targets ← do match genSelPats with
| none => pure <| iHypsToGeneralize hyps fvar
| some genSelPats =>
  let genSelTargets ← SelPat.resolve hyps genSelPats
  checkDependentHyps hyps genSelTargets
  let explicitIrisTargets := genSelTargets.filter (match ·.kind with | .ipm _ => true | _ => false)
  let explicitPureTargets := genSelTargets.filter (match ·.kind with | .pure _ => true | _ => false)
  let implicitIrisTargets := (iHypsToGeneralize hyps fvar).filter (not <| (explicitIrisTargets.map (·.kind)).contains ·.kind)
  pure <| explicitPureTargets ++ implicitIrisTargets ++ explicitIrisTargets
...
```
Any thoughts?

alvinylt:
There are still two syntactic features available in Lean's built-in `induction` tactic but are not yet implemented for `iinduction`.

The first is the support for multiple induction targets (`elimTarget,+`). The syntax of the built-in `induction` tactic is defined in [Init.Tactics](https://github.com/leanprover/lean4/blob/master/src/Init/Tactics.lean#L1001-L1002):
```lean4
syntax (name := induction) "induction " elimTarget,+ (" using " term)?
  (" generalizing" (ppSpace colGt term:max)+)? (inductionAlts)? : tactic
```
This is unlike the current `iinduction` syntax which handles one term at a time:
```lean4
syntax (name := iinduction) "iinduction" colGt term
    ("using" ident)?
    ("generalizing" (ppSpace colGt selPat)+)?
    (inductionAlts)? : tactic
```

The second is to support `term` instead of `ident` for the recursor name under the `using ...` clause.

Other than these two, I think all syntactic features with the built-in `induction` tactic have also been implemented for `iinduction`.

As one can tell from the current `iinduction` implementation, there is some reuse of syntax definitions from [`Init.Tactics`](https://github.com/leanprover/lean4/blob/master/src/Init/Tactics.lean). Meanwhile, there is some syntax parsing done with the function `parseInductionAlts`, which resembles some code in [`Lean.Elab.Tactic.Induction`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Tactic/Induction.lean) but is not reusable for some reasons (e.g., requires intertwining with `ProofModeM`-related functions). Also, are the parsing functions in [`Lean.Elab.Tactic.Induction`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Tactic/Induction.lean) (e.g., `evalAlt`) not accessible externally because they are not within a `public` section? We might need to think of some workaround for this, but I also reckon there should be priority for testing `iinduction` and using it in IPM proofs, which I'm hoping to finish within the upcoming few days.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved, see: https://github.com/leanprover-community/iris-lean/pull/430#issuecomment-4644138453

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved, see: https://github.com/leanprover-community/iris-lean/pull/430#issuecomment-4644062295

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: I added a new function in `ProofModeM.lean` named `addBIGoalRunTactics` for this purpose.

Iris/Iris/Tests/Tactics.lean
alvinylt: Tests added in commits [107e19d](https://github.com/leanprover-community/iris-lean/commit/107e19d6606a4bb0f8f9d96e589157cf43e87f38) and [57ecb81](https://github.com/leanprover-community/iris-lean/commit/57ecb8112977d843bcee0ff0bb3e9f717fdbed2a).

alvinylt:
Here is a summary of the progress in the last couple of days.

### Fixing issue with `irevert`

I believe there is an issue with `irevert`. Here is an example:

```lean4
example [BI PROP] {P : PROP} {m n : Nat} {Q R S : Nat → PROP} {T : Nat → Prop}
    {h1 : T m} {U1 : (T m) → Prop} {h2 : U1 h1} :
    ⊢@{PROP} True := by
  irevert %m %h1 %U1 %h2
```

`irevert` gives the following error:
```
Tactic `clear` failed: target depends on 'h1'
```

I think this has to do with how `RevertState.revertLeanHyp` handles pure Lean hypothesis. In particular, a pure Lean hypothesis is universally quantified only when it is a `PROP`. However, in the example above, the pure hypothesis `h2` depends on `h1`, which in turn depends on `m`.

Now that it is fixed in commit [9ba7098](https://github.com/leanprover-community/iris-lean/commit/9ba70988f3f64d892ab776c2240404a0834d316e), `h2` is reverted as a wand premise while `m`, `h1` and `U1` are reverted as universally quantified variables.

This fixes the tactic failure when similar sets of hypotheses are reverted using the `generalizing` syntax with `iinduction`.

### "Try this" suggestion for `irevert` and `iloeb`

The "Try this" suggestion is implemented also for `irevert` and `iloeb`. The function `checkDependentHyps`, which checks for forward-dependent hypotheses, is moved to [`Revert.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInduction/Iris/Iris/ProofMode/Tactics/Revert.lean#L102-L198) and slightly generalised.

Try this suggestion with irevert

Try this suggestion with iloeb

### Using `iinduction` in existing inductive proofs

There are several proofs that use induction within the Iris Proof Mode. Many other proofs under `BI/` and `Instances/` are done by induction but not within IPM. So I guess `iinduction` is not relevant in those cases.

Commits: [17aefd6](https://github.com/leanprover-community/iris-lean/pull/430/commits/17aefd6fe2995c06870298fc5d67d91fe1ca1371), [f1f0263](https://github.com/leanprover-community/iris-lean/pull/430/commits/f1f02636e789afee3f97f2972231694ff29ba218), [2d26cef](https://github.com/leanprover-community/iris-lean/pull/430/commits/2d26cef9dff8a0bc6bab825d418c84ea96e98984)

alvinylt:
### More Tests

More tests for `iinduction` are added into [`Tests/Tactics.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInduction/Iris/Iris/Tests/Tactics.lean#L2805-L3092). This includes tests with lists, natural numbers and binary trees. There are also tests to ensure the tactic catches errors (duplicate cases, missing cases, etc.) and throws the correct messages.

### Testing `IntoIH` type class instances: `intoIH_listForall` and `intoIH_listForall₂`

Lean does not seem to generate induction principles for nested/mutual inductive datatypes. Instead they have been defined manually:

https://leanprover.zulipchat.com/#narrow/channel/113489-new-members/topic/.E2.9C.94.20Induction.20principle.20for.20nested.20inductive.20types/near/437905021

We can define a simple function that mirrors an `NTree` and prove a lemma about it:
```lean4
def NTree.mirror : NTree α → NTree α
  | .leaf => .leaf
  | .node x ts => .node x (ts.map NTree.mirror)

example [BI PROP] {α} {t : NTree α} :
  ⊢@{PROP} ⌜.mirror (.mirror t) = t⌝ := by ...
```

Because of how the induction principle is defined, the induction hypothesis for the step case is `∀ x, <affine> ⌜x ∈ ts⌝ -∗ ⌜x.mirror.mirror = x⌝`, so no `List.Forall` involved. Note that [the Rocq version](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/tests/proofmode.v) (lines 2588-2614) defines the induction principle for `ntree` that explicitly uses `List.Forall` manually instead of using that generated by Rocq. 

Actually, it seems like [`List.Forall₂`](https://github.com/leanprover-community/batteries/blob/main/Batteries/Data/List/Basic.lean#L378-L388) is under `Batteries.Data.List` while [`List.Forall`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/List/Defs.html#List.Forall) live in `Mathlib.Data.List.Defs` and thus requires an import. Instead, there is [`List.all`](https://github.com/leanprover/lean4/blob/master/src/Init/Data/List/Basic.lean#L1913-L1926) (under `Init.Data.List`), which outputs a `Bool` value instead of a `Prop`. The `IntoIH` type class instance [`intoIH_listForall`](https://github.com/ISTA-PLV/iris-lean/blob/iInduction/Iris/Iris/ProofMode/Instances.lean#L949-L969) in this PR is defined in terms of `List.all`, so it is technically different from the original formulation in Rocq. Should we keep it or reformulate it?

The type class instance `intoIH_listForall₂` is less of a concern as `List.Forall₂` exists in the Lean. A binary relation [`NTree.Rel`](https://github.com/ISTA-PLV/iris-lean/blob/iInduction/Iris/Iris/Tests/Tactics.lean#L2884-L2922) is defined along with its induction principle (for the latter, by Claude).

Iris/Iris/ProofMode/Instances.lean
alvinylt: Further comments: https://github.com/leanprover-community/iris-lean/pull/430#issuecomment-4712950236

Iris/Iris/ProofMode/Instances.lean
alvinylt: Further comments: https://github.com/leanprover-community/iris-lean/pull/430#issuecomment-4712950236

alvinylt:
[#batteries > Chain and Chain' and Pairwise @ 💬](https://leanprover.zulipchat.com/#narrow/channel/348111-batteries/topic/Chain.20and.20Chain.27.20and.20Pairwise/near/540472211):

> `List.Forall` deliberately doesn't exist, you are supposed to use `∀ x ∈ l, p x`

I guess in this case `intoIH_forall` is already sufficient?

One could really have defined an induction principle using `List.all` like this:

```lean4
def NTree.induction_principle {α : Type}
    (p : NTree α → Prop) [DecidablePred p]
    (h_leaf : p leaf)
    (h_node : (x : α) → (ts : List (NTree α)) →
      List.all ts (fun t => decide (p t)) = true → p (node x ts)) :
    ∀ t : NTree α, p t := ...
```

I assume one would not want to do this, as we get induction principles such as:
```
IH : (ts.all fun t => decide (⊢ ⌜t.mirror.mirror = t⌝)) = true
```
which nonetheless simplifies to `∀ (t : NTree α), t ∈ ts → ⊢ ⌜t.mirror.mirror = t⌝` with `simp_all`.

Commit: [c684029](https://github.com/leanprover-community/iris-lean/pull/430/commits/c6840290af3091319c9c629c9477d5b42d3ee4c3)

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2775,4 +2783,310 @@ example (P Q : PROP) :
...
+/--
+  Definition of n-tree and its induction principle from:
+  https://leanprover.zulipchat.com/#narrow/channel/113489-new-members/topic/.E2.9C.94.20Induction.20principle.20for.20nested.20inductive.20types/near/437905021
+-/
+inductive NTree (α : Type)
+| leaf
+| node : α → List (NTree α) → NTree α
+
+@[induction_eliminator]
+def NTree.induction_principle {α} (p : NTree α → Prop) (h_leaf : p leaf)
```
MackieLoeffel: Please add a test that uses this induction principle.

Iris/Iris/ProofMode/Tactics/Revert.lean
```diff
@@ -106,7 +230,8 @@ def iRevertCore (targets : List SelTarget) {u : Level}{prop: Q(Type $u)}{bi : Q(
   let pf' : Q($(st.e) ⊢ $(st.goal)) ← withoutFVars (u:=0) st.reverted (k st.hyps st.goal)
   return q($(st.pf) $pf')
 
-elab "irevert" pats:(colGt selPat)+ : tactic => do
+elab_rules : tactic
+  | `(tactic| irevert $pats:selPat*) => do
```
MackieLoeffel, line 260: Is this change necessary?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,517 @@
...
+  hypotheses in the Iris goal to be generalised. This includes the subset of
+  hypotheses in the intuitionistic context with occurrences of `fvar`, as well
+  as all spatial hypotheses.
+
+  These hypotheses are then reverted from the the Iris contexts into the regular
+  Lean context before applying Lean's built-in `induction` tactic.
+-/
+private def iHypsToGeneralize {u} {prop : Q(Type $u)} {bi} {e : Q($prop)}
+    (hyps : Hyps bi e) (fvar : FVarId) : List SelTarget :=
+  let ivars := hyps.intuitionisticIVarIds.filter <| fun ivar =>
```
MackieLoeffel: Why are the intuitionistic hypotheses implicitly generalized instead of giving an error like for pure hypotheses?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,517 @@
...
+  induction target `n`.
+  By applying `iinduction n generalizing HQ`, the hypotheses `HQ`
+  are additionally reverted and thus included as premises in the induction
+  hypothesis.
+  One can also generalise pure variables in the regular Lean context. However,
+  if there exists some another pure/Iris hypothesis that is forward-dependent.
+  For example, `iinduction n generalizing %R` is not valid as `HR` depends on `R`.
+  Instead, one can use `iinduction n generalizing %R HR`.
+-/
+elab_rules : tactic
```
MackieLoeffel, line 380: Why is this using `elab_rules` instead of `elab`?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,517 @@
...
+            let ihFVars ← findIHs s.mvarId
+
+            -- A proof that all hypotheses in the Iris goal exist in the intuitionistic context
+            let some pfIntHyps := irisGoal.hyps.buildIntuitionisticProof
+            | throwOrLogError "iinduction: spatial context should be empty"
+
+            -- Introduce the induction hypothesis back into the Iris proof state
+            let st ← addIHs pfIntHyps irisGoal.hyps ihFVars
+
+            match parsedAlts with
```
MackieLoeffel: It seems like there is code duplicated between the different branches. Can this be avoided?

MackieLoeffel (commented):
I looked over the code and overall it seems very nice, but there is also a lot of code and I need more time to review it more in depth.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: This follows the tactic behaviour in Rocq: intuitionistic hypotheses that depends on the induction target are generalised automatically by the tactic, while intuitionistic hypotheses that depend on any pure hypothesis in the `generalizing` clause must also be explicitly generalised by the tactic user.

For example:
```rocq
Lemma test_induction_1 (m n : nat) (predicate : nat → PROP) P (Q : nat → PROP) (R : nat → PROP) S T :
  P -∗ □ Q n -∗ □ R m -∗ S -∗ □ T -∗ ⌜(n + 0 = n)%nat ⌝.
Proof.
  iIntros "HP #HQ #HR HS #HT".
  iInduction n as [].
Abort.
```
The induction hypothesis is `"IHn" : □ Q n -∗ P -∗ S -∗ ⌜(n + 0)%nat = n⌝`.

If we generalise the pure variable `m`, the tactic throws an error stating that `HR` depends on `m`:
```rocq
Lemma test_induction_2 (m n : nat) (predicate : nat → PROP) P (Q : nat → PROP) (R : nat → PROP) S T :
  P -∗ □ Q n -∗ □ R m -∗ S -∗ □ T -∗ ⌜(n + 0 = n)%nat ⌝.
Proof.
  iIntros "HP #HQ #HR HS #HT".
  iInduction n as [] forall (m).
Abort.
```

The solution is to have `HR` generalised as well:
```rocq
Lemma test_induction_3 (m n : nat) (predicate : nat → PROP) P (Q : nat → PROP) (R : nat → PROP) S T :
  P -∗ □ Q n -∗ □ R m -∗ S -∗ □ T -∗ ⌜(n + 0 = n)%nat ⌝.
Proof.
  iIntros "HP #HQ #HR HS #HT".
  iInduction n as [] forall (m) "HR".
Abort.
```
The induction hypothesis is then `∀ m0 : nat, □ Q n -∗ P -∗ S -∗ □ R m0 -∗ ⌜(n + 0)%nat = n⌝`.

In fact, pure hypothesis dependencies are handled in the same way. Consider the following example where `h` depends on the induction target `n`:
```rocq
Lemma test_induction_4 (m n : nat) (predicate : nat → Prop) (h : predicate n) P :
  P -∗ ⌜(n + 0 = n)%nat ⌝.
Proof.
  iIntros "HP".
  iInduction n as [].
Abort.
```
It follows that the induction hypothesis is `<affine> ⌜predicate n⌝ -∗ P -∗ ⌜(n + 0)%nat = n⌝`.

Suppose `h` now depends on `m`, and we would generalise like to `m`.
```rocq
Lemma test_induction_5 (m n : nat) (predicate : nat → Prop) (h : predicate m) P :
  P -∗ ⌜(n + 0 = n)%nat ⌝.
Proof.
  iIntros "HP".
  iInduction n as [] forall (m).
Abort.
```
The tactic then throws an error about the dependency, which can be easily be solved by using the tactic `iInduction n as [] forall (m h)` instead.

The Lean implementation follows the same design.

Iris/Iris/Tests/Tactics.lean
alvinylt: An example is added in commit [d2dfc2f](https://github.com/leanprover-community/iris-lean/pull/430/commits/d2dfc2f160a32f6fd63b9c07670cdab9b3da8096).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: So here's how I understand the tactic in Rocq:
1. Intuitionistic hypotheses dependent on the *induction target* are implicitly generalised.
2. Pure/intuitionistic hypotheses dependent on any hypothesis in the `generalizing` clause must also be included in the `generalizing` clause explicitly.

Actually there might possibly be an issue in the Rocq implementation.
Here's an example:

```rocq
Lemma test_induction_6 (n : nat) (pred : nat → Prop) (H : pred n) (R : pred n → PROP) :
  □ R H -∗ True.
Proof.
  iIntros "#HR".
  iInduction n as [].
Abort.
```

The type class synthesis with `IntoIH` fails:
```
Tactic failure: iInduction: cannot import IH
(∀ (H : pred n) (R : pred n → PROP),
"HR" : R H
--------------------------------------□
True) into proof mode context (IntoIH instance missing).
```

By the current design `HR` should probably be implicitly reverted as it transitively depends on `n`. But [`iHypsContaining` in `ltac_tactics.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/ltac_tactics.v#L1877-L1889) does not compute the transitive closure.

This issue is easily fixed in the Lean version in commit [0724695](https://github.com/leanprover-community/iris-lean/pull/430/commits/0724695071f9acac541b589511a3fb9dd4cd2912) since `collectForwardDeps` is a function available in Lean's library. Now `iHypsToGeneralize` finds all pure hypotheses that are dependent on the induction target and stipulates the reverting of intuitionistic hypotheses that depend on any of the said pure hypotheses.

Perhaps we might also like to fix this in the Rocq version?

I reckon the most elegant approach is either to have all dependencies automatically handled by the tactic or to always stipulate the use of `generalizing` in all cases. But I can also understand the benefit of the current design in Rocq, which distinguishes induction target dependencies from dependencies within the `generalizing` clause.

Iris/Iris/ProofMode/Tactics/Revert.lean
alvinylt: I think so if we would like to use `checkDependentHyps` for `irevert`.

The syntax of the tactic is first declared so that we can specify what `checkDependentHyps` should suggest:
```lean4
syntax (name := irevert) "irevert" (colGt ppSpace selPat)+ : tactic
```

In commit [0724695](https://github.com/leanprover-community/iris-lean/pull/430/commits/0724695071f9acac541b589511a3fb9dd4cd2912), I have refactored the relevant code, so hopefully this will be easier to read.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Same reason as in https://github.com/leanprover-community/iris-lean/pull/430#discussion_r3428182031.

The syntax of `iinduction` is declared at the top of `Induction.lean`. Furthermore, there are also functions for syntax parsing in `Induction.lean` which (as far as I know) necessitates a separate syntax declaration.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Yes, I have refactored the code for the continuation function `k'` and `withoutFVars (u := 0) ...`.

I've also removed some unnecessary variable declarations and pattern matching on `Option` types by using `mapM` and `forM`.

Commits: [2721180](https://github.com/leanprover-community/iris-lean/pull/430/commits/2721180c95998963ce204271d62b087d3185b939), [65b92b3](https://github.com/leanprover-community/iris-lean/pull/430/commits/65b92b3e1c434b6412f941d3b835b4e2db69a366), [3bd9128](https://github.com/leanprover-community/iris-lean/pull/430/commits/3bd91283fdd3ca9635d66e3b9c2e434d1bc07056), [2cbe8ef](https://github.com/leanprover-community/iris-lean/pull/430/commits/2cbe8efe850b82b43191065031e2d9354c4f989e), [9c4fc02](https://github.com/leanprover-community/iris-lean/pull/430/commits/9c4fc02526e53d6816095378ec42ef2ad85fbcb3)

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+      throwOrLogErrorAt alt.stx
+        m!"iinduction: duplicate alternative name `{alt.ctor}`"
+
+    handledCtors := handledCtors.insert alt.ctor
+
+private def iInductionCore {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e}
+    (hyps : Hyps bi e) (goal : Q($prop)) (fvar : FVarId)
+    (parsedAlts : Option Alts)
+    (altRecName : Option Name)
+    (genSelTargets : Option <| List SelTarget) :
```
MackieLoeffel: Remove the `Option`.

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+  -- Find the regular pure/Iris hypotheses to be reverted
+  let implicitIrisTargets ← iHypsToGeneralize hyps fvar
+  let targets ← do
+    match genSelTargets with
+    | none => pure implicitIrisTargets
+    | some genSelTargets =>
+      let explicitIrisTargets := genSelTargets.filter (match ·.kind with | .ipm _ => true | _ => false)
+      let explicitPureTargets := genSelTargets.filter (match ·.kind with | .pure _ => true | _ => false)
+      let implicitIrisTargets := implicitIrisTargets.filter (not <| (explicitIrisTargets.map (·.kind)).contains ·.kind)
+      pure <| explicitPureTargets ++ implicitIrisTargets ++ explicitIrisTargets
```
MackieLoeffel: Don't do a reorder, but just use the order the user specified (plus the spatial hyps at the end).

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+
+/-- Search for induction hypotheses in the pure Lean context -/
+private def findIHs (m : MVarId) : ProofModeM (List FVarId) :=
+  m.withContext do
+    let lctx ← getLCtx
+    let mut ihs := []
+    for decl in lctx do
+      let type ← instantiateMVars decl.type
+      -- The hypothesis is an IH when it is an Iris entailment
+      if containsIrisGoal type.getForallBody then
```
MackieLoeffel: Replace with `isIrisGoal type.getForallBody.getAppFn`?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+    match hyps.getDecl? ivar with
+    | some ⟨_, _, _, ty⟩ => dependentFVars.any (ty.containsFVar ·)
+    | none => false
+
+  return (ivars ++ hyps.spatialIVarIds).map ({ kind := .ipm ·, explicit := false })
+
+/-- Search for induction hypotheses in the pure Lean context -/
+private def findIHs (m : MVarId) : ProofModeM (List FVarId) :=
+  m.withContext do
+    let lctx ← getLCtx
```
MackieLoeffel: Maybe the `findIHs` function can be replaced by a one-liner like `getLCtx.filterM (λ decl => do type ← instantiateMVars decl.type; return isIrisGoal type.getForallBody.getAppFn).reverse`?

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+            -- For pretty printing of arguments
+            parsedAlts.forM fun parsedAlts => do
+              (parsedAlts.alts.find? (matcher ctor) <|> parsedAlts.wildcard).forM
+                fun ⟨_, vars, _, stx⟩ => do
+                  if vars.size > s.fields.size then
+                    throwOrLogErrorAt stx <|
+                      s!"iinduction: too many variable names provided at alternative `{ctor}`: ".append
+                      s!"{vars.size} provided, but {s.fields.size} expected"
+
+                  -- For pretty printing of arguments
```
MackieLoeffel: pretty print -> hover information

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+          | _ => Name.mkSimple "_") }
+      | none =>
+        -- Defer the check if the first tactic for the `with` syntax is given
+        if parsedAlts.tac.isSome then
+          pure { explicit := true, varNames := [] }
+        else
+          throwMissingAlt ctor
+
+  let pf ← iRevertIntro hyps goal targets <|
+    fun hyps' goal' k => do
```
MackieLoeffel: `k` -> `k_intro`

Iris/Iris/ProofMode/Tactics/Induction.lean
```diff
@@ -1,59 +1,442 @@
...
+                      s!"{vars.size} provided, but {s.fields.size} expected"
+
+                  -- For pretty printing of arguments
+                  for ⟨fieldFVar, varStx⟩ in s.fields.toList.zip vars.toList do
+                    if let `(binderIdent| $id:ident) := varStx then
+                      let lctx ← getLCtx
+                      let fieldType ← inferType fieldFVar
+                      addLocalVarInfo id lctx fieldFVar (some fieldType) true
+
+            let k' : ProofModeContinuationIntro := fun hyps goal => do match parsedAlts with
```
MackieLoeffel: inline it so we don't need a name

Iris/Iris/ProofMode/ProofModeM.lean
```diff
@@ -96,6 +96,46 @@ def addMVarGoal (m : MVarId) (name : Name := .anonymous) : ProofModeM Unit := do
...
+    pure (m, none, subgoals)
+  | some firstTactic =>
+    let subgoals ← evalTacticAt firstTactic m.mvarId!
+    let mut subgoals' := []
+    -- Run the user tactics on each newly generated goal after running `firstTactic`
+    for s in subgoals do
+      subgoals' := subgoals'.append <| ← evalTacticAt tacticSeq s
+    -- Add the result proof subgoals into the proof state
+    for s' in subgoals' do addMVarGoal s'
+    pure (m, subgoals, subgoals')
```
MackieLoeffel: Deduplicate with something like
```
-- run first tactic
let subgoals := match firstTactic with | none => [m.mvarId!] | some firstTactic => evalTacticAt firstTactic m.mvarId!

-- run tacticSeq
let mut subgoals' := []
-- Run the user tactics on each newly generated goal after running `firstTactic`
for s in subgoals do
  subgoals' := subgoals'.append <| ← evalTacticAt tacticSeq s
-- Add the result proof subgoals into the proof state
for s' in subgoals' do addMVarGoal s'
pure (m, subgoals, subgoals')
```

Iris/Iris/ProofMode/ProofModeM.lean
```diff
@@ -96,6 +96,46 @@ def addMVarGoal (m : MVarId) (name : Name := .anonymous) : ProofModeM Unit := do
...
+  The function returns:
+  1. the metavariable of the initial goal,
+  2. the metavariables of subgoals after running the first tactic, and
+  3. the metavariables of subgoals after running the remaining tactic sequences.
+-/
+def addBIGoalRunTactics {prop : Q(Type u)} {bi : Q(BI $prop)}
+    {e} (hyps : Hyps bi e) (goal : Q($prop)) (name : Name := .anonymous)
+    (firstTactic : Option <| TSyntax `tactic)
+    (tacticSeq : TSyntax `Lean.Parser.Tactic.tacticSeq) :
+    ProofModeM (Q($e ⊢ $goal) × Option (List MVarId) × List MVarId) := do
```
MackieLoeffel: simplify this to return a bool if the first tactic succeeded.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [0a6fd19](https://github.com/leanprover-community/iris-lean/pull/430/commits/0a6fd190af87eff2d799737164cf1a6e3e51c950).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [40141f3](https://github.com/leanprover-community/iris-lean/pull/430/commits/40141f388f557572d1a73a9e479a4cb7499d4065).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [7df0c36](https://github.com/leanprover-community/iris-lean/pull/430/commits/7df0c36ca214e389f8ca9a5416fb20eb51a18ccb).

Iris/Iris/ProofMode/ProofModeM.lean
alvinylt: Resolved in commit [2cbd1ad](https://github.com/leanprover-community/iris-lean/pull/430/commits/2cbd1ad011823179d4c18ed231e2b42efaadca0f).

Iris/Iris/ProofMode/ProofModeM.lean
alvinylt: Resolved in commit [2cbd1ad](https://github.com/leanprover-community/iris-lean/pull/430/commits/2cbd1ad011823179d4c18ed231e2b42efaadca0f).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [40141f3](https://github.com/leanprover-community/iris-lean/pull/430/commits/40141f388f557572d1a73a9e479a4cb7499d4065).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commits [7afbc84](https://github.com/leanprover-community/iris-lean/pull/430/commits/7afbc84554e3d74b3c4c430c5c4bae64c7088bee) and [2943d7a](https://github.com/leanprover-community/iris-lean/pull/430/commits/2943d7a4813719d919eedcbbbf9659772dfdce28). Since `findIHs` is used only once within `iInductionCore`, it is moved inline.

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: `getAppFn` returns only the first expression (e.g. `List.Forall₂`) rather than the arguments. There is `getAppArgs`, which returns the arguments as an array. However, even `type.getForallBody.getAppArgs.any isIrisGoal` would not be fully correct as the Iris entailment may be hidden under several layers of application.

Fortunately, the built-in libraries already provide `Expr.find?` which recursively checks whether a predicate (in our case `isIrisGoal`) holds anywhere in the expression. This means `containsIrisGoal` can be simplified as follows.
```lean4
def containsIrisGoal (e : Expr) : Bool := (e.find? isIrisGoal).isSome
```
Since `containsIrisGoal` is used only once within `iInductionCore`, it is further eliminated by moving the code inline.


Commit: [89a8226](https://github.com/leanprover-community/iris-lean/pull/430/commits/89a82269bb9f6f03d4f257031f3f5c913e4a3c95)

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in commit [e54087d](https://github.com/leanprover-community/iris-lean/pull/430/commits/e54087d24856b8afb0b471016ed867ed0a419eb3).

Iris/Iris/ProofMode/Tactics/Induction.lean
alvinylt: Resolved in https://github.com/leanprover-community/iris-lean/pull/430#discussion_r3435366311.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2775,4 +2789,388 @@ example (P Q : PROP) :
...
+• Lean hypothesis `h1` depends on `n`
+• Lean hypothesis `x` (inaccessible name) depends on `n`
+• Iris hypothesis `x` (inaccessible name) depends on `n`
+-/
+#guard_msgs in
+example [BI PROP] [BILoeb PROP] {n : Nat} {P T : Nat → PROP} {Q : Nat → Prop}
+    {h1 : Q n} {_ : (Q n) → Prop} :
+    ⊢ □ T n -∗ □ P n := by
+  iintro #_
+  iloeb as IH generalizing %n
```
MackieLoeffel, line 2833: Please add a test that uses iloeb with the `!` syntax and the same for irevert

alvinylt: Added in commits [e91f09e](https://github.com/leanprover-community/iris-lean/pull/430/commits/e91f09eebdd5c9a9943f875861dedc8c564b6e09) and [348b093](https://github.com/leanprover-community/iris-lean/pull/430/commits/348b09378e7f97024f144951ea4b90716244a83b).

MackieLoeffel:
Thanks for all the revisions! I looked over the changes and I think we can merge this. Please resolve the merge conflict and I will merge this MR.

alvinylt:
Thanks for the review! The merge conflict has now been now resolved.

MackieLoeffel:
Great, thank you! Merged.

Issue #431, opened 2026-06-01
Port iris_heap_lang/proofmode.v


Kaptch:
Depends on:
- #283 
- #284

lzy0505:
Tracked by #480

PR #412, opened 2026-05-26
feat: GenHeap and HeapLang example

ayhon:
## Description
A `GenHeap` implementation as a thin wrapper around `GhostMap`. Since `reservation_map` is not defined yet, we chose a prior version's [`gen_heap.v` file](https://gitlab.mpi-sws.org/iris/iris/-/blob/61840d318d6e147f3d8b8bb635b7235ee6d9c501/theories/base_logic/lib/gen_heap.v) to port.

Depends on #407, fixes #283.

## Checklist
* [ ] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [ ] I have added my name to the `authors` section of any appropriate files

PR #434, opened 2026-06-01
feat: `wp_bind`

ayhon:
## Description
Implementation of the `wp_bind` tactic.

Depends on #412, addresses #431.

## Checklist


Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -0,0 +1,96 @@
...
+  let K ← elabTermEnsuringTypeQ K q(HeapLang.Exp)
+  trace[bind] s!"Context to bind over: {←ppExpr K}"
+  ProofModeM.runTactic fun mvar {u, prop, bi, hyps, goal, ..} => do
+    let .defEq _ ← isLevelDefEqQ u 0
+      | throwError "`wp_bind` only works over `IProp` (at universe level 0)"
+    let ~q(IProp $GF) := prop
+      | throwError "`wp_bind` only works over `IProp`"
+    let ~q(UPred.instBIUPred) := bi
+      | throwError "`wp_bind` expected the BI implementation of `IProp` to be `UPred.instBIUPred`"
+
```
MackieLoeffel, line 52: I think it would be great to have a function `iWpBindCore` that mirrors `reshape_expr` for Iris Rocq. I am imagining a signature like the following (rough sketch)
```
def iWpBindCore  {GF} {e} (hyps : Hyps (IProp $GF) e) (goal : Q($prop))
  (k : (e : Q(Exp)) → (Φ : _) → ProofModeM (Option Q(WP $e  {{ $Φ }})) ) : ProofModeM Q($e ⊢ $goal)
```
The idea would be that `k` would be called for all expressions `e` in an evaluation position and `k` can then either return a proof of a WP, which is then lifted to a proof of the WP in the evaluation context, or it can return none, in which case the next context is tried.

Having such a function would make it quite easy to implement all the different tactics that need binding like `wp_bind`, `wp_pure`, and `wp_smart_apply`.

ayhon: Hmm, I feel like if we massage the signature enough, we might be able to make it fit with [`ForIn.forIn`'s](https://leanprover-community.github.io/mathlib4_docs/Init/Core.html#ForIn.forIn). Even if not, I think using `ForInStep` instead of `Option` would make sense, to allow early stops

MackieLoeffel: Good idea, this would be really nice!

MackieLoeffel: Or alternatively to directly using `ForIn`, maybe the function should return an [iterator](https://lean-lang.org/doc/reference/latest/Iterators/#--tech-term-iterator) ([docs](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Iterators/Basic.html#Std.Iterator))?

lzy0505:
#412 has been merged. Can you rebase onto master?

Iris/Iris/HeapLang/ProofMode.lean
ayhon: I think to mimic Rocq, the signature should be more like
```
def iWpBindCore  {GF} {e} (hyps : Hyps (IProp $GF) e) (goal : Q($prop))
  (k : (K : Q(List ECtxItem))(e : Q(Exp)) → (Φ : _) → ProofModeM (Option Q(WP fill $K $e  {{ $Φ }})) ) : ProofModeM Q($e ⊢ $goal)
```
I think this may have been what you meant originally, but I spent some trying thinking how the prior signature would work, by taking the `P |- WP e {{ Φ }}` proof and lifting it to some `P |- WP fill K e {{ Φ' }}`, but then I need to relate `Φ` with `Φ'`, which is a bit more complicated that the signature above.

I think the signature I proposed would work well with `ForIn` / `iterator`s. This is what I have so far, in a continuation style
```lean4
def runOverAllEvaluationContexts  {GF : Q(BundledGFunctors.{0, 0, 0})} -- TODO: Why explicit universes needed?
  (k : ∀ {Δ : Q(IProp $GF)}{inst : Q(Wp (IProp $GF) Exp Val Stuckness)} (s : Q(Stuckness))(E : Q(CoPset))(K : Q(List ECtxItem))(e : Q(Exp)) (Φ : Q(Val → IProp $GF)),
     ProofModeM (Option Q($Δ ⊢ Wp.wp $s $E (fill $K $e) $Φ)))
  {Δ : Q(IProp $GF)}
  (hyps : Hyps q(inferInstance) Δ)
  (goal : Q(IProp $GF))
  (focus? : Option Q(Exp)) :
    ProofModeM Q($Δ ⊢ $goal) := do
```

ayhon: I have reworked this so it works by iterating over a call to `extractAllOuterEvCtx e`, which gives all pairs `(K, e')` such that `fill K e' = e`. For `wp_bind` it suffices to pick the first one which matches the `focus`, for `wp_pure` iterate over all options and `continue` if no `PureExec` (or `IntoLaterN`) instances are found.

ayhon:
There are some invariants in this code that I couldn't enforce at compile time. In particular, I have my own implementation of `fill` and `fillItem` at the meta level, that I must assert are equivalent for Qq to typecheck. I experimented with having these functions also provide a proof of correctness, but:
 - When building a proof term `q(e = fill K e')` I ran into some issues with metavariables hidden by Qq which I spent too much time trying to solve. (Also, these proofs in the end were defeq to `rfl`, so it seemed wasteful to create a huge proof term equivalent to `rfl`. This is what made me attempt the next point)
 - Trying to use Qq assertions `$e =Q fill $K $e'` as invariants was not any better, since these assertions were not checked at compile-time (at which point, it's equivalent to having one big assertion at the end) and couldn't be combined to produce more facts (this is because, I believe, these assertions rely on syntactic equality, not definitional equality, which is what I actually needed)
 
I am not terribly happy with having basically two implementations of `fill`, the one given by `EctxLanguage` at the object level and the one I just implemented at the meta level. The advantage of this division is that it's not necessary to perform a simplification pass on the expression afterwards. For `wp_bind` this presents "some" advantage, but for `wp_pure`, since we want to process the resulting goal either way to apply `wp_value` if necessary, the advantage is not that clear. See [the `wp_pure` PR](https://github.com/leanprover-community/iris-lean/blob/c5050eb433b370ce61199fbfa13783771590d493/Iris/Iris/HeapLang/ProofMode.lean#L54-L109) for an alternate solution relying on invoking `simp only`

ayhon:
One discrepancy with Rocq's implementation is that my `extractAllOuterEvCtx` does not correspond precisely with Rocq's `reshape_expr`, due to how `match` backtracking works in the Rocq implementation. In particular, `reshape_expr` when invoked on `#1 + #2` will consider both `fill [#1 + ·] #2` and `fill [· + #2] #1`, whereas my implementation does not give `fill [#1 + ·] #2`. 

I argue that this discrepancy is fine, because all ignored cases are those for which `e'` is a value, for which it doesn't make much sense to use `wp_bind` (because why would you want to have a `WP` over a value) nor `wp_pure` (because a value doesn't reduce). However, I wanted to document this here in case someone came up with a counterexample of why we would want to recover this behavior. It might even be relevant to document this in the code as well.

MackieLoeffel:
I think it is easier to review this together with the wp_pure tactic in #437 and close this PR. @ayhon Does this work for you?

ayhon:
> I think it is easier to review this together with the wp_pure tactic in #437 and close this PR. @ayhon Does this work for you?

Sure, that's reasonable. I can do it in 15 mins, or feel free to merge this into the other PR yourself

MackieLoeffel:
Is it necessary to do a merge? It looks like all the commits from this PR are already in #437

ayhon:
> Is it necessary to do a merge? It looks like all the commits from this PR are already in #437

Aah, I was thinking of merging #437 into #434, to preserve the comments I have here (also, the branch is called `feat/basic-wp-tactics`, so I feel it fits better, although it doesn't really matter)

MackieLoeffel:
That is a good point, then let's do that. I can handle that.

ayhon:
> That is a good point, then let's do that. I can handle that.

I just got pushed it, sorry!

MackieLoeffel:
I guess you were faster :) I closed #437 in favor of this PR.

Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -0,0 +1,180 @@
...
+    e = ProgramLogic.fill K e' →
+    Φ' = (λ v : Val => WP (ProgramLogic.fill K (Exp.val v)) @ s; E {{ Φ }}) →
+    (Δ ⊢ WP e @ s ; E {{ Φ }})
+    := by
+  rintro H rfl rfl
+  iintro _
+  iapply wp_bind
+  iapply H $$ [$]
+
+elab "wp_bind" focus:hl_exp : tactic => do
```
MackieLoeffel: I made a pass over `wp_bind` and cleaned up the implementation. Maybe these changes help to also clean up `wp_pure`.

MackieLoeffel: The notation for `wp_bind` is broken. You will see if you actually try to use it in a proof. The notation for `wp_pure` has similar problems.

MackieLoeffel: Please update Spinlock.lean to use wp_bind and wp_pure to make sure that they work.

Iris/Iris/Tests/HeapLang/WeakestPre.lean
```diff
@@ -0,0 +1,123 @@
...
+error: unsolved goals
+hlc : HasLC
+GF : BundledGFunctors
+ι : IrisGS_gen hlc Exp GF
+⊢ ⏎
+  ⊢ WP hl(((#0 + #1) + #2)) {{ v, WP hl((v(v) + #3)) {{ v, True }} }}
+-/
+#guard_msgs in
+example : ⊢@{IProp GF}  WP hl(((#0 + #1) + #2) + #3) {{ v, True }} := by
+  wp_bind ((#0 + _) + _)
```
MackieLoeffel, line 78: These tests are good, but please also add tests that finish the proof to make sure that the proof terms are constructed correctly.

Iris/Iris/HeapLang/ProofMode.lean
ayhon: I think the problem comes from `elabTermEnsuringTypeQ` not actually running macro expansion. I think this could be fixed by having ``macro "wp_apply" focus:hl_expr : tactic => `(tactic| wp_apply hl($focus))``.

ayhon: I'm running into issues with the metavariable in `focus` not being assigned in `isDefEq`. The hover info provided by `trace[..] m!"{focus}"` tells me that the metavariable is synthetic (which would make sense with how `isDefEq` doesn't assign it), but inspecting manually gives me it's `.natural`. I'll investigate further in the morning.

Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -0,0 +1,172 @@
...
+    let newGoal := q(Wp.wp $s $E $inner $Φ)
+
+    let .some ⟨Δ'₂, hyps'⟩ := parseHyps? bi <| ←instantiateMVars Δ'
+      | throwError s!"Obtained hypothesis {←ppExpr Δ'} from `IntroLaterN`, but these couldn't be parsed as {←ppExpr <| ←inferType bi} hypothesis"
+      -- We make use of the invariant that `parseHyps? Δ = .some (Δ', hyps) → Δ = Δ'`.
+    have : $Δ'₂ =Q $Δ' := ⟨⟩
+    let nextPf ← addBIGoal hyps' newGoal
+
+    let HΦ : Q($φ) ← mkFreshExprSyntheticOpaqueMVar q($φ)
+    iSolveSideconditionAt HΦ.mvarId! (failOnUnsolved := false)
```
ayhon: Is `iSolveSideconditionAt` enough? I had chosen `grind only` because it would also take care of running `lia`.

Iris/Iris/HeapLang/ProofMode.lean
ayhon: Found the issue. I was elaborating the term outside the `mvar.withMainContext` ran by `ProofModeM.runTactic`. Idk why this didn't make it fail in some files though, I'd like to investigate further

Iris/Iris/HeapLang/ProofMode.lean
MackieLoeffel: Rocq uses [solve_vals_compare_safe](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris_heap_lang/proofmode.v?ref_type=heads#L155) which is defined [here](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris_heap_lang/proofmode.v?ref_type=heads#L133), so no fancy solver. Any unsolved sidecondition will be given to the user so they can call grind after the tactic if necessary.

Iris/Iris/HeapLang/ProofMode.lean
MackieLoeffel: No, I don't think this is the problem.
I think the problem comes from
```
wp_bind x
iapply l
```
being parsed as
```
wp_bind x iapply l
```
where `focus` is instantiated with `x iapply l`. 
I added a `colGt` to the app notation of HeapLang and now it works.

Iris/Iris/HeapLang/Lib/SpinLock.lean
```diff
@@ -117,6 +118,8 @@ theorem try_acquire_spec (γ : GName) (lk : Val) (R : IProp GF) :
   icases Hlock with ⟨%l, %Heq, #Hinv⟩
   subst Heq
   iapply wp_bind (fun x => hl(snd({x})))
+  -- TODO: why does the following not work?
+  -- wp_bind snd(_)
```
MackieLoeffel: Somehow the isDefEq to find the right subexpression fails here and I have no idea why. @ayhon Do you have an idea what the problem could be?

ayhon: I believe this is the issue I addressed above, where we're elaborating `focus` outside of `ProofModeM.runWpTactic`. I have a fix in my local version, I have just been doing a bunch of cleanup and haven't pushed it yet. Give me 10 mins

MackieLoeffel: You are right, I missed your comment above. Good spot, I don't know if I would have found the cause of this problem!

ayhon:
Ok, I pushed my PR addressing all previous issues. `SpinLock` uses `wp_bind` and works correcly. I also pushed a bunch of the `=Q` Qq assertions to come from the functions which guarantee them, instead of asserting them at the call site (the idea being that it's easier for users to discover them this way). I also dropped the `Option` in the signature of `finECtx` and opted for making an instance of `MonadBacktrack ProofModeM`. That way we no longer need to use `withoutModifyingState` in `wp_pure` and we can make use of the `Alternative` methods to fail if necessary.

Lmk what you think!

MackieLoeffel:
Looks great! I will make a pass over the PR now.

Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -0,0 +1,154 @@
...
+  H.trans (wp_value_fupd ⟨rfl⟩).2
+
+public theorem tac_wp_value_nofupd [ι : IrisGS_gen hlc Exp GF] {Δ} {s : Stuckness} {E : CoPset} {v : Val} {Φ : Val → IProp GF}
+  (H : Δ ⊢ Φ v) :
+  (Δ ⊢ WP (v : Exp) @ s ; E {{ Φ }}) :=
+  H.trans <| fupd_intro.trans (wp_value_fupd ⟨rfl⟩).2
+
+elab "wp_value_head" : tactic =>
+  ProofModeM.runTacticWp fun mvar {GF, hyps, s, E, e, Φ, ..} => do
+    let ~q(ProgramLogic.ToVal.ofVal $v) := e
```
MackieLoeffel: This code here should be extracted into a function that is then used at the end of wp_pure. @ayhon Let me know if you will do this, otherwise I can also do it later.

MackieLoeffel (commented):
I am done with my pass. I added a wp_value_head tactic that should be integrated with wp_pure, but I have to run now. Otherwise this is ready to merge I think. Thanks for all your great work on this @ayhon !

Iris/Iris/HeapLang/ProofMode.lean
ayhon: I'm back on my keyboard and happy to do this.

ayhon: I'm running into issues on how to define the signature of this function. If I take the fields of `IrisGoal` directly, I cannot make the return type typecheck, because (it seems) Qq doesn't use the `=Q` assertions.
```lean4
public meta
def iWpValueHead
  {u : Level}
  {prop : Q(Type u)}
  {bi : Q(BI $prop)}
  {ehyps : Q($prop)}
  (hyps : Hyps bi ehyps)
  {GF : Q(BundledGFunctors.{0, 0, 0})}
  {hlc : Q(HasLC)}
  (ι : Q(IrisGS_gen $hlc Exp $GF))
  (s : Q(Stuckness))
  (E : Q(CoPset))
  (e : Q(Exp))
  (Φ : Q(Val → IProp $GF))
  (hu : (u =QL 0) := ⟨⟩)
  (hprop : $prop =Q IProp $GF := ⟨⟩)
  (hbi : $bi =Q UPred.instBIUPred := ⟨⟩) :
    ProofModeM Q($ehyps ⊢ Wp.wp $s $E $e $Φ) := do
```

However, if I ask for `WpGoal` I cannot express the return type of the function either, because one also cannot access those assertions.

I tried "baking in" those assertions into the signature directly, simplifying a lot of stuff, but that still gives me the following error:
image

I think the only option might be to have this function return a `Lean.Expr`. I'm going to try to arrive to a better compromise, but what do you think about that?

ayhon: I ended up adding an extra `Wp $prop Exp Val Stuckness` parameter, and requiring it to be `wp.def`.

MackieLoeffel:
I made a final pass over the tactics and everything looks good now. Thanks for all your work on this!

PR #435, opened 2026-06-01
fix: make IProp universe polymorphic

MackieLoeffel:
Currently `IProp` is hardcoded to universe-level `Type 0`. I don't see why we need this and changing it to be universe polymorphic seems to work. @markusdemedeiros Do you know if it is necessary to force `IProp` in `Type 0`?


markusdemedeiros:
I do not think it is necessary, this seems like a mistake

markusdemedeiros:
It seems that `Countable` and `ConstOF` are also pinned at universe 1. However, generalizing `constOF` seems to be pretty bad. I'll make a new issue for that.

MackieLoeffel:
Sounds good, thanks for having a look. I am merging this.

PR #437, opened 2026-06-02
feat: `wp_pure` tactic

ayhon:
## Description
Implements the `wp_pure` tactic.

Depends on #434, addresses #431.

## Checklist


Iris/Iris/Tests/HeapLang/WeakestPre.lean
```diff
@@ -0,0 +1,123 @@
...
+-- TODO: Why is this not an `instance`?
+attribute [local instance] snd_pure
+
+/--
+error: unsolved goals
+hlc : HasLC
+GF : BundledGFunctors
+ι : IrisGS_gen hlc Exp GF
+⊢ ⏎
+  ⊢ WP hl(#0) {{ v, True }}
```
Kaptch, line 92: Would it make sense to also apply `wp_val`, when the expression reduces to a value? I think there's no loss of information, but honestly I don't remember how it works in the Rocq version.

ayhon: Actually, I think the Rocq implementation does attempt to use `wp_val`, by passing the new goal to `wp_finish`. This is something I missed in my implementation.

ayhon: If some kind of postprocessing of the goal is something we'll need to do anyways, then maybe having (the currently, horribly named) `HeapLang.fill` is not really necessary, since one can just use the object level `ProgramLogic.fill` and then invoke `simp` with some custom lemmas (or the new `cbv`)

MackieLoeffel:
Closing in favor of #434

PR #441, opened 2026-06-03
Quicksort

MackieLoeffel:
This PR adds a Quicksort example that I plan to use as part of the demo at the Iris Workshop. This PR also contains a lot of cleanup:

- Add precedence levels to the pointsto notation such that one can omit parentheses
- allow ∀ x : X notation for BI ∀
- allow escaping for hl_binder
- change wp lemmas to have generic postcondition
- add a wp_rec tactic for unfolding a recursive function at the start of a proof 
- consistently use ToVal.ofVal (aliased to Exp.ofVal) instead of Exp.val
- add a lemma to abstract common parts of Language.PureExec proofs
- use ?x notation for escaping terms in heaplang, replacing `{term}` with `&(term)`
- add `wp_expr_simp` and `wp_finish` tactics and automatically use them from `wp_pure`
- clean up parsing and printing of literals


Iris/Iris/HeapLang/Lib/Quicksort.lean
```diff
@@ -0,0 +1,241 @@
...
+def cons : Val := hl_val% λ x, some(ref(x))
+
+def append : Val := hl_val%
+  rec append x :=
+    let l1 := fst(x);
+    let l2 := snd(x);
+    match l1 with
+    | none() => l2
+    | some(x) =>
+      (let p := !x;
```
markusdemedeiros: Given that match is normally in tail position (at least in the HeapLang programs I've written) perhaps we change the precedence of the match arms to remove these parens?

markusdemedeiros:
I do think it's a little unfortunate that the `?name` syntax clashes with the regular syntax for metavariables. Maybe `&name`? Some type of upwards pointing arrow (though many of those clash with casts anyways)? This is a total bikeshed though, the current syntax is a definite improvement over `{...}`.

MackieLoeffel:
I also don't like `?` very much, but I don't know what a better character is. `&`, `%`, `*` all don't work because these operators are already used for binary operators. e.g. if we use `&`, `a &b` could either mean `.and (.var a) (.var b)` or `.app (.var a) b`. `$` does not work since it is already used for antiquotations. `#` is used for literals. `@` might work, but that clashes with making all implicit parameters explicit. `:` or `\` might work? `!` is used for load.

Iris/Iris/HeapLang/Lib/Quicksort.lean
```diff
@@ -0,0 +1,242 @@
...
+      iapply cons_spec $$ Hl2
+      iintro %_ _
+      wp_pures
+      imodintro
+      iapply HΦ $$ Hl1
+      have : x < hd := by grind
+      simp [*]
+      iframe
+
+-- set_option profiler true in
```
markusdemedeiros: Is this here for the purposes of the demo? Same with the commented out examples below?

Iris/Iris/HeapLang/Lib/Quicksort.lean
```diff
@@ -5,7 +5,7 @@ public import Iris.HeapLang.ProofMode
 
 namespace Iris.HeapLang
 
-open BI Iris ProgramLogic
+open BI Iris ProgramLogic List
```
MackieLoeffel, line 8: My understanding of the Mathlib convention is to not open the `List` namespace, but use `List.Perm` and so on instead. @markusdemedeiros Do you know what the convention here is.

Iris/Iris/HeapLang/Lib/Quicksort.lean
MackieLoeffel: This one can be removed. The comments below I want to keep since they show the pure conditions that are solved by grind. This might be useful for the demo.

MackieLoeffel:
I just realized that if we use the Lean notation for bitwise operators (i.e. `&&&` instead of `&` and so on), we can both use `&` for the embedding and fix the precedence of match. I've done this in https://github.com/leanprover-community/iris-lean/pull/441/commits/2280225d67c4527eed4986a702b8f286eb522125

Iris/Iris/HeapLang/Lib/Quicksort.lean
markusdemedeiros: I think open `List` is fine (at least they do it all over the place, and all it does here is make the intermediate states of the proof easier to parse). Did you see differently somewhere?

markusdemedeiros:
As expected, this is good to go! Just made the instances a little easier to read. Overall I will say, this is a super impressive example of our (your) quality of life features. It will make for a very good demo.

Iris/Iris/HeapLang/Lib/Quicksort.lean
MackieLoeffel: Fixed as described in https://github.com/leanprover-community/iris-lean/pull/441#issuecomment-4634929540

MackieLoeffel:
Thanks for the review! I did a final cleanup and merged.

PR #443, opened 2026-06-03
refactor: replace Frac hierarchy with Rat 

markusdemedeiros:
## Description
re. https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Delete.20Frac.20typeclass/with/599809140
Work in progress

## Checklist

## Generative AI Guidelines


Iris/Iris/Algebra/Lib/FracAuth.lean
```diff
@@ -25,88 +25,92 @@ open Iris OFE CMRA UCMRA Auth Option
...
 notation "◯F " a => fragFull a
 
-abbrev fracOne : Frac F := ⟨One.one⟩
+abbrev fracOne : Qp := One.one
 
-instance frac_one_exclusive (b : A) : Exclusive (fracOne (F := F), b) where
-  exclusive0_l y h := absurd h.1 (not_exists.mp UFraction.one_whole.2 y.1.1)
+instance frac_one_exclusive (b : A) : Exclusive (fracOne, b) where
+  exclusive0_l y h := by
+    have hy := y.1.2
```
markusdemedeiros: Golf me

Iris/Iris/Algebra/HeapView.lean
```diff
@@ -343,9 +343,15 @@ theorem update_one_delete :
       obtain ⟨_, _, _, Hqv, Hinc⟩ := Hrel
       have Hval := Option.validN_of_incN_validN (Hv := Hqv) (Hinc := Hinc)
       rcases v with ⟨(f|_|f), _⟩
-      · exact one_whole.2 ⟨f, Hval.1⟩ |>.elim
-      · exact one_whole.2 Hval.1 |>.elim
-      · exact one_whole.2 ⟨f, Fraction.Fractional.proper Hval.1⟩ |>.elim
+      · exact absurd Hval.1
```
markusdemedeiros: Golf me

markusdemedeiros:
Requested reviews from Sergei, Zongyuan and Michael: I think the changes should be relatively uncontroversial, but I want to know your guys' opinion on at least A) if we should stick to the current Rocq mirror of having a a Qp type, or alternatively just use a raw type of Rat and encode positivity into the validity predicates to maybe get better automation, and B)  if the tactic code that builds off of fractions (like icombine) should be changed at all in lieu of this switch

Iris/Iris/Algebra/Lib/FracAuth.lean
```diff
@@ -25,88 +25,91 @@ open Iris OFE CMRA UCMRA Auth Option
...
-public abbrev fragFull (a : A) : FracAuth (F := F) (A := A) := frag ⟨One.one⟩ a
+public abbrev fragFull (a : A) : FracAuth (A := A) := frag One.one a
 
 notation "●F{" dq "} " a => auth dq a
 notation "●F " a => auth (DFrac.own One.one) a
 notation "◯F{" q "} " a => frag q a
 notation "◯F " a => fragFull a
 
-abbrev fracOne : Frac F := ⟨One.one⟩
+abbrev fracOne : Qp := One.one
```
MackieLoeffel: Is this abbrev still necessary or can we just use `1`?

Iris/Iris/Algebra/Lib/MonoNat.lean
```diff
@@ -37,45 +37,43 @@ scoped instance : CMRA.CoreId (a : MaxNat) := OrdCommMonoidLike.instCoreId _
 
 end MaxNat
 
-abbrev MonoNat (F : Type _) [UFraction F] := Auth F MaxNat
+abbrev MonoNat := Auth MaxNat
```
MackieLoeffel, line 40: Question independent of this PR: Why is this an abbrev? I would expect this to be a `def`.

Iris/Iris/Algebra/DFrac.lean
```diff
@@ -97,136 +89,129 @@ instance DFrac_CMRA : CMRA (DFrac F) where
...
-    · exact UFraction.one_whole.2 ⟨_, Hv1⟩
-    · exact UFraction.one_whole.2 Hv1
-    · exact UFraction.one_whole.2 <| Fractional.of_add_left Hv1
+    have h1 : (One.one : Qp).val = 1 := rfl
+    rcases y1 with (y|_|y) <;>
+      simp only [CMRA.ValidN, CMRA.op, op, valid] at Hv1 <;>
+      first | (have := y.2; grind) | grind
 
-instance one_exclusive_right [CMRA V] {v : V} : CMRA.Exclusive (v, own (F := F) One.one) where
+instance one_exclusive_right [CMRA V] {v : V} : CMRA.Exclusive (v, own (One.one : Qp)) where
```
MackieLoeffel, line 129: Should this be `One.one` or `1`?

Iris/Iris/Algebra/Frac.lean
```diff
@@ -13,238 +13,120 @@ public import Iris.Algebra.IsOp
...
-  Proper : α → Prop
-  add_comm : ∀ a b : α, a + b = b + a
-  add_assoc : ∀ a b c : α, a + (b + c) = (a + b) + c
-  add_left_cancel : ∀ {a b c : α}, a + b = a + c → b = c
-  /-- There does not exist a zero fraction. -/
-  add_ne : ∀ {a b : α}, a ≠ b + a
-  proper_add_mono_left : ∀ {a b : α}, Proper (a + b) → Proper a
+/-- The type of positive rational numbers, used as fractions -/
+@[rocq_alias frac, rocq_alias fracO, rocq_alias fracR]
+abbrev Qp := { q : Rat // 0 < q }
```
MackieLoeffel, line 26: Should this be a `def` instead of an `abbrev`? Also have you thought about making this a structure instead of using `Subtype`?

Iris/Iris/Algebra/Frac.lean
```diff
@@ -13,238 +13,120 @@ public import Iris.Algebra.IsOp
...
+  comm := OFE.leibniz.mpr <| Subtype.ext (Rat.add_comm ..)
+  pcore_op_left H := by rcases H
+  pcore_idem H := by rcases H
+  pcore_op_mono H := by rcases H
+  extend {_ x y z} := by rintro H rfl; exists y; exists z
+
+
+-- TODO: A different solution to having these bridge lemmas might be to internalize
+-- positivity into the CMRA's validity predicate, removing the sybtype, and having Qp
+-- become just a Leibniz CMRA over Rat. This admits two-way coercions to Rat for the automation.
```
MackieLoeffel, line 73: I think the current approach makes sense. But I would leave this comment and maybe one can experiment with this in the future.

MackieLoeffel (commented):
I think this PR is great and simplifies a lot of things. It is a bit inconsistent about using `One.one` vs `1`. It might be worth doing a search and replace to clean that up.

lzy0505:
The PR looks good.

> A) if we should stick to the current Rocq mirror of having a a Qp type, or alternatively just use a raw type of Rat and encode positivity into the validity predicates to maybe get better automation

I like staying close to Rocq. Based on my experience, I don’t see many practical use cases for Qp automation. Do you have anything concrete in your mind?

markusdemedeiros:
> I like staying close to Rocq. Based on my experience, I don’t see many practical use cases for Qp automation. Do you have anything concrete in your mind?

linarith for the fractions. Like right now we need a custom wrapped + operation to package the result with a positivity proof, instead of Rat's +. This manifests as this big ad-hoc block of simp and grind lemmas around line 74 of frac.

I'll consider it acceptable for now but keep a comment marking the choice

Iris/Iris/Algebra/Frac.lean
markusdemedeiros: I was hoping that using a subtype instead of an equivalent but custom would assist `simp` and `grind`. Tbh I think this is better unless you have another thought

Iris/Iris/Algebra/Frac.lean
```diff
@@ -13,238 +13,120 @@ public import Iris.Algebra.IsOp
...
+
+@[simp, grind =] theorem Qp.val_add (x y : Qp) : (x + y).val = x.val + y.val := rfl
+@[simp, grind =] theorem Qp.val_one : (1 : Qp).val = 1 := rfl
+@[simp, grind =] theorem Qp.val_half (q : Qp) : q.half.val = q.val / 2 := rfl
```
Shreyas4991, line 77: This should be automatic from a simps annotation for `def half`. Unfortunately simps is currently in mathlib.

Shreyas4991 (commented):
I think you can get a lot of structure projection simp lemmas that you prove via rfl by tagging structure instances with `@[simps]`. Unfortunately this is currently a mathlib tactic. It is worth requesting upstreaming it, since the pattern appears to repeat in CMRAs. 

Simps docstring page here : https://leanprover-community.github.io/mathlib4_docs/Mathlib/Tactic/Simps/Basic.html?utm_source=chatgpt.com


(Apologies for getting the link through gpt, but it works and sadly GPT searches mathlib docs better than Google).

Iris/Iris/Algebra/Lib/FracAuth.lean
markusdemedeiros: It seems to interpret it as a Nat when we do this instead of Qp. Not exactly sure why, there are a couple weird cases like this. I'm going to merge and then make this a new issue.

PR #463, opened 2026-06-13
feat: add missing PureExec instances

MackieLoeffel:
Generated by Opus 4.8

### DIFF
--- a/Iris/Iris/HeapLang/Instances.lean
+++ b/Iris/Iris/HeapLang/Instances.lean
@@ -170,10 +170,36 @@ instance PureExec_pair {v1 v2 : Val} : Language.PureExec True 1 hl((&v1, &v2)) h
       intro Ki e_inner heq
       cases Ki <;> cases heq <;> simp [toVal]
 
-instance PureExec_le_int {n1 n2 : Int} : Language.PureExec True 1 hl(#(.int n1) ≤ #(.int n2)) hl(#(.bool (n1 ≤ n2)))  where
-  pureExec _ := by
+set_option synthInstance.checkSynthOrder false in
+instance instPureExecUnOp {op : UnOp} {v v' : Val} :
+    Language.PureExec (op.eval v = some v') 1 (Exp.unop op (.ofVal v)) (.ofVal v') where
+  pureExec h := by
     refine .once <| mk_pure_prim_step (fun _ => ?_) (fun _ _ _ _ _ hs => ?_) ?_
-    · constructor <;> simp [BinOp.eval]
+    · constructor <;> simp [*]
+    · cases hs <;> simp_all [UnOp.eval]
+    · apply EctxItemLanguage.subredexes_are_values
+      intro Ki e_inner heq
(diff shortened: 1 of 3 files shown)


Iris/Iris/HeapLang/Instances.lean
```diff
@@ -170,10 +170,36 @@ instance PureExec_pair {v1 v2 : Val} : Language.PureExec True 1 hl((&v1, &v2)) h
       intro Ki e_inner heq
       cases Ki <;> cases heq <;> simp [toVal]
 
-instance PureExec_le_int {n1 n2 : Int} : Language.PureExec True 1 hl(#(.int n1) ≤ #(.int n2)) hl(#(.bool (n1 ≤ n2)))  where
-  pureExec _ := by
+set_option synthInstance.checkSynthOrder false in
```
markusdemedeiros, line 173: I'm wondering if it might be better to not have a general lemma for all `op.eval`, and instead have a specialized version for each constructor, to avoid `set_option synthInstance.checkSynthOrder false`. What do you think?

Iris/Iris/HeapLang/Instances.lean
```diff
@@ -170,10 +170,36 @@ instance PureExec_pair {v1 v2 : Val} : Language.PureExec True 1 hl((&v1, &v2)) h
...
+  pureExec h := by
+    refine .once <| mk_pure_prim_step (fun _ => ?_) (fun _ _ _ _ _ hs => ?_) ?_
+    · constructor <;> simp [*]
+    · cases hs <;> simp_all [BinOp.eval]
+    · apply EctxItemLanguage.subredexes_are_values
+      intro Ki e_inner heq
+      cases Ki <;> cases heq <;> simp [toVal]
+
+-- higher priority than the generic binop instance
+instance (priority := default + 10) instPureExecEqOp {v1 v2 : Val} :
```
markusdemedeiros, line 197: Do you know why this is necessary? I suspect that the instance priority modifier could go away if we specialized binop as well but I'm not 100% clear on that.

Iris/Iris/HeapLang/Instances.lean
MackieLoeffel: These instances use the ipm TC synthesis and with that there is no problem with the synth order. So I think the warning is fine. There are 14 binary operations, so I think it is better to not repeat this instance 14 times. This is also how it is handled in Rocq.

Iris/Iris/HeapLang/Instances.lean
MackieLoeffel: It is a more specialized instance than the binop instance, so this specialized instance should take priority. This is the same as in Rocq.

Iris/Iris/HeapLang/Instances.lean
markusdemedeiros: Ah right. Thanks for the explanation!

MackieLoeffel:
Thanks for having a look! Merging.

PR #483, opened 2026-06-23
feat: coercion from Lean.Name to Namespace

ayhon:
## Description
There seems to be a quite direct mapping from `Lean.Name` to `Namespace`. This PR implements a coercion from the latter to the former, to allow writing `Namespace` literals using Lean's `Name` literal syntax.

Before:
```lean4
def N : Namespace := nroot.@"oneshot"
```

Now:
```lean4
def N : Namespace := `oneshot
```

## Checklist

### DIFF
--- a/Iris/Iris/Std/Namespaces.lean
+++ b/Iris/Iris/Std/Namespaces.lean
@@ -28,6 +28,13 @@ def nclose (N : Namespace) : CoPset :=
 
 instance : CoeOut Namespace CoPset where coe := nclose
 
+def ofName : Lean.Name → Namespace
+  | .anonymous => nroot
+  | .str rest str => ndot (ofName rest) str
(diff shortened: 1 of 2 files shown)


markusdemedeiros:
This is kind of cool, lol. Out of curiosity, do you think we could eliminate the string version?

ayhon:
I don't think we can really phase it out, since the rest of the file defines theorems for nroot and ndot which I assume are needed elsewhere. We could make the notation local fo the file though, to encourage people to use the Name coercion

MackieLoeffel:
People sometimes do something like `N.@l` where `N` is a namespace and `l` is a location. I guess for this we cannot use the fancy new syntax but need the old version.

markusdemedeiros:
Merging this

PR #485, opened 2026-06-25
Leibniz OFE construction

Kaptch:
## Description

POC: OFE constructor that takes:
- Carrier: `X: Type _`;
- Distance: `dist: Nat \to X \to X \to Prop`, satisfying old distance axioms.

And returns a OFE that has `X/\sim` carrier, where `x \sim y \iff \forall n. dist n x y` and lifted distance. 
It trivially satisfies old `equiv_dist` axiom, and is Leibniz. 
Tested it on agreement, and it was relatively fast to fix downstream problems (~10 lines in total). 

## Updated summary
Made Agree a Leibniz OFE, on Agree A, equiv is now just =.

There is a generic construction in OFE.lean (QuotientO + mkQuotient, plus a little mk/lift/map wrapper API), then a Leibniz version of Agree on top of it. Agree API is the same, except that now it requires its type argument to be OFE even for the definition.
Once Agree is Leibniz, a bunch of cameras built on it are too, so I added the instances: View, Auth, and the lib ones (DFracAgree, ExclAuth, FracAuth). Plus two tiny fixes where the refactor changed how Agree values get constructed.
The only non-Leibniz (or Leibniz-preserving) OFEs/CMRAs are related to maps, and I'll do them in a separate PR shortly.

### DIFF
--- a/Iris/Iris/Algebra/Agree.lean
+++ b/Iris/Iris/Algebra/Agree.lean
@@ -264,6 +264,12 @@ instance instRawSetoid : Setoid (Raw α) :=
 /--
 EXPERIMENT: Leibniz Agree.
 https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Evaluating.20a.20specialization.20to.20Leibnize.20OFE.27s/with/606745235
+
+Note that the Rocq version of Agree does not require that the agreement type be an OFE. The
+typeclass addition here is trivial: if your [α] type was _not_ an OFE then you should instanstiate
+it with the trivial one. If your type already has a _different_ OFE, and you want agreement
+over the equality relation rather than your OFE equivalence, you should introduce a wrapper type
+with the trivial OFE.
 --/
 @[rocq_alias agree]
 def Agree (α : Type u) [OFE α] : Type u := Quotient (instRawSetoid (α := α))
--- a/Iris/Iris/Algebra/Agree.lean
+++ b/Iris/Iris/Algebra/Agree.lean
@@ -28,7 +28,7 @@ OFE equivalence makes `≡` coincide with `=`, so `Agree α` is `OFE.Leibniz` by
 
 /-! ## The raw representation -/
 
-section raw
+section Raw
 
 variable {α : Type u} {β : Type v}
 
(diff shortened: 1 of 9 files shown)


markusdemedeiros:
Nice! Any reason why this touches views/agree?

Kaptch:
I tested it on agreement, because it is one of a few algebras, which would be changed, if we move to a new definition of OFEs without equiv. Views were a collateral damage, there was minor case of looking past the interface of agreement.  
But apart from minor nits in agreement-dependent algebras, nothing changes much. We never eliminate from algebras to Lean types, so there's no need for Quotient.lift in other parts of the project.
Morally, this example shows that changing equiv -> eq would not break agreement algebra, since we could provide the same interface. 
I haven't tested it with other algebras with non-trivial equiv, but it should work fine.

markusdemedeiros:
That's promising!

Iris/Iris/Algebra/OFE.lean
```diff
@@ -248,6 +248,49 @@ theorem Equiv.to_eq {α} [OFE α] [Leibniz α] {x y : α} (h : x ≡ y) : x = y
   ⟨eq_of_eqv, .of_eq⟩
 export OFE.Leibniz (leibniz)
 
+/-- The setoid on `X` identifying points that agree at every step index:
+`x ≈ y ↔ ∀ n, dist n x y`. -/
+def distSetoid {X : Type u} (dist : Nat → X → X → Prop) (heqv : ∀ n, Equivalence (dist n)) :
```
Shreyas4991: `Setoid` is a type class.  We typically use `instance` instead of `def`.

Iris/Iris/Algebra/Agree.lean
```diff
@@ -143,202 +165,340 @@ theorem Agree.op_ne {x : Agree α} : OFE.NonExpansive x.op := by
...
+variable {α : Type u} {β : Type v} [OFE α] [OFE β]
+
+instance instRawSetoid : Setoid (Raw α) :=
+  QuotientO Raw.dist fun {_} => Raw.dist_equiv
+
+/--
+EXPERIMENT: Leibniz Agree.
+https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Evaluating.20a.20specialization.20to.20Leibnize.20OFE.27s/with/606745235
+
+Note that the Rocq version of Agree does not require that the agreement type be an OFE. The
```
markusdemedeiros: @Kaptch I'm a little confused as to why the type needs to be an `OFE` here, when the type did not need to be an `OFE` before. Isn't the point of the `Agree` quotient just to make it a set rather than a list?

markusdemedeiros (changes requested):

Iris/Iris/Algebra/Agree.lean
Kaptch: Under [OFE \alpha] it should be isomorphic to Agree := { X : FiniteSet \alpha | X \neq \empty }
The reason I used this implementation is that I thought that making our version of FiniteSet seems to be redundant, given that it will be used only in Agree, and Mathlib has a version of FiniteSet with good API, so porting it would introduce duplication of one more chunk of Mathlib.
This version of Agree can be used with non-OFE types by giving them discrete metric.
Also, it was a good usecase for a more general quotient construction, which could be reused for some client OFEs if they also required implicit quotienting via Equiv. 
That being said, I can redo it with finite sets instead, I have a simple version of FiniteSet somewhere, and can use that.

markusdemedeiros: I think you succeeded in showing that the general quotient construction works, but yes, it sounds like we should definitely avoid the extra OFE constraint.

Kaptch: Oki, I will think of a good way to do it then.

Kaptch: Instead of bringing finsets for one definition, quotiented lists in-place, and restored the old interface without the OFE obligation.

markusdemedeiros: Sounds great. Re-request when I can review it!

markusdemedeiros (approved):
Looks superb!

lzy0505 (approved):

PR #487, opened 2026-06-28
feat: iaccu

### LINKED ISSUE #364: Port iAccu

alvinylt:
## Description

Implements the `iaccu` tactic.

Addresses #364.

The function `Hyps.buildAccuProof` directly builds the proof required by `iaccu` from a `Hyps` instance. Since the set of hypotheses is organised as a tree, a simple recursion on a `Hyps` instance would have resulted in the spatial hypotheses being combined with arbitrary associativity. Instead, `Hyps.buildAccuProofAux` ensures that the combined proposition is right-nested.

## Checklist


Iris/Iris/ProofMode/Expr.lean
```diff
@@ -189,6 +189,35 @@ partial def Hyps.intuitionisticIVarIds {u prop bi} :
   | _, .hyp _ _ ivar p _ _ => if isTrue p then [ivar] else []
   | _, .sep _ _ _ _ lhs rhs => lhs.intuitionisticIVarIds ++ rhs.intuitionisticIVarIds
 
+private def Hyps.buildAccuProofAux {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e e' : Q($prop)}
```
MackieLoeffel: Instead of using an explicit `...Aux` definition, it is usually better to use a `where` clause to make clear that this is an auxiliary function and should not use used directly. See e.g. https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/Tactics/ModIntro.lean#L101

MackieLoeffel (commented):
Except for one small comment, this PR looks good to me.

Iris/Iris/ProofMode/Expr.lean
alvinylt: Resolved in commit [1b94af9](https://github.com/leanprover-community/iris-lean/pull/487/commits/1b94af9c55c0692613b1b7a3b1656cb187aad33b).

MackieLoeffel:
Thanks, merged!

PR #491, opened 2026-07-01
doc: update ProofMode.Display comment

ayhon:
## Description
The comment in `ProofMode.Display` references an implementation of the display which is no longer used. This PR updates the comment.

We may consider making this a module doc comment.


MackieLoeffel:
Good spot, thanks!

PR #496, opened 2026-07-02
feat: remaining introduction patterns and case destruction patterns

alvinylt:
## Description

Implements the remaining introduction patterns and case destruction patterns.

Addresses #213 and #239.

Features that are only used with `iintro` on the top level are implemented as introduction patterns in [`IntroPattern.lean`](https://github.com/ISTA-PLV/iris-lean/blob/IntroPats/Iris/Iris/ProofMode/Patterns/IntroPattern.lean) and [`Intro.lean`](https://github.com/ISTA-PLV/iris-lean/blob/IntroPats/Iris/Iris/ProofMode/Tactics/Intro.lean).

| Feature in Rocq             | Corresponding constructor in `IntroPat` | Remarks                                                                                                                                                                                                                                                                                                                                                          | Ported? |
|-----------------------------|-----------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
| IForall (`*`)               | `all` (`*`)                             | Introduces all universal quantifiers.                                                                                                                                                                                                                                                                                                                            | ✓       |
| IAll (`**`)                 | `allwand` (`**`)                        | Introduces all universal quantifiers, pure arrows and wands.                                                                                                                                                                                                                                                                                                     | ✓       |
| IPureIntro (`!%`)           | `pureintro` (`!%`)                      | Introduces a pure proof goal, which is basically the same as running `ipureintro`. This does not necessarily have to be the last in the introduction pattern sequence. If there are further introduction patterns, attempt to enter IPM again and handle the rest of the patterns. This means applying `iintro !% %_ !%` to `⊢@{PROP} ⌜⌜ϕ⌝ ⊢@{PROP} ⌜ϕ⌝⌝` works. | ✓       |
| ISimpl (`/=`)               | `simp` (`/=`)                           | Apply `dsimp`. Using `dsimp` instead of `simp` here so that it resembles the use of `simpl` in Rocq. To solve the goal directly, one can use `//=` instead (see below).                                                                                                                                                                                          | ✓       |
| IDone (`//=`)               | `simptrivial` (`//=`)                   | Shorthand for applying `/=` and then `//`.                                                                                                                                                                                                                                                                                                                       | ✓       |
[...]

PR #470, opened 2026-06-16
feat: iinv

alvinylt:
## Description

Implements the `iinv` tactic.

Addresses #317 and #363.

This includes porting the following type classes instances of `IntoAcc`, `ElimAcc`, `IntoInv` and `ElimInv`.

| Type class instance          | Located in ... (Rocq)                                                                                                       | Located in ... (Lean)                                                                                                | Definition ported? | Proof ported? |
|------------------------------|-----------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------|--------------------|---------------|
| `elim_inv_acc_without_close` | [`class_instances.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances.v)                  | [`InstancesIris.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInv/Iris/Iris/ProofMode/InstancesIris.lean)       | ✓                  | ✓             |
| `elim_inv_acc_with_close`    | [`class_instances.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances.v)                  | [`InstancesIris.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInv/Iris/Iris/ProofMode/InstancesIris.lean)       | ✓                  | ✓             |
| `elim_acc_bupd`              | [`class_instances_updates.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_updates.v)  | [`InstancesUpdates.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInv/Iris/Iris/ProofMode/InstancesUpdates.lean) | ✓                  | ✓             |
| `elim_acc_fupd`              | [`class_instances_updates.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_updates.v)  | [`InstancesUpdates.lean`](https://github.com/ISTA-PLV/iris-lean/blob/iInv/Iris/Iris/ProofMode/InstancesUpdates.lean) | ✓                  | ✓             |
[...]

PR #464, opened 2026-06-14
doc: tactic descriptions with docstrings

alvinylt:
## Description

Currently some IPM tactics are not annotated with comments. As a result, when one hovers over a tactic, only a generic pop-up message is shown.

Generic pop-up message 1

With docstrings added for each `elab`/`macro`, we get more specific descriptions about the IPM tactics. This resembles what we get with built-in tactics in Lean and hopefully makes IPM nicer to use.

Nice pop-up message with a description of the IPM tactic

I wrote these comments manually based on [`tactics.md`](https://github.com/leanprover-community/iris-lean/blob/update_tactics_md/Iris/tactics.md) and examples from [`Tactics.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/Tests/Tactics.lean).

If there are some notable differences in terms of syntax or tactic behaviour compared to the Rocq tactics, I think it might also be helpful to highlight them in these comments. Would be great to hear whether others think this would be useful.

## Checklist
* [x] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [x] I have added my name to the `authors` section of any appropriate files

Issue #400, opened 2026-05-20
Porting missing instances for WP


Kaptch:
#393 introduced WP. At the time, some ProofMode typeclasses were missing. When the PM starts supporting those, the missing instances should be ported as well.

- [x] AddModal
- [x] ElimAcc #470
- [x] FrameInstantiateExistDisabled #511

alvinylt:
The `ElimAcc` instances are being ported in #470 and will be available along with `iinv` once they get merged.

PR #500, opened 2026-07-05
feat: remaining specialisation patterns

alvinylt:
## Description

Implements the remaining specialisation patterns.

Addresses #215 and #239.

| Feature in `SpecPat`                                                   | Remarks                                                                                                                                                                                                                                                                                                                                                                                                                                  | Ported? |
|------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
| Nested specialisation patterns (`.ident` with specialisation patterns) | The functions `processWand` and `iSpecializeCore` are now mutual definitions. Similarly, `pmTerm` and `specPat` are defined by mutual induction, so the definitions in [`ProofModeTerm.lean`](https://github.com/ISTA-PLV/iris-lean/blob/master/Iris/Iris/ProofMode/Patterns/ProofModeTerm.lean) are moved into [`SpecPattern.lean`](https://github.com/ISTA-PLV/iris-lean/blob/SpecPats/Iris/Iris/ProofMode/Patterns/SpecPattern.lean). | ✓       |
| Subgoal for persistent premise (`.goal` with `kind := .persistent`)    | For the specialisation pattern `[# $H₁ … $Hₙ ]`, where `H₁ … Hₙ` are framed. Every hypotheses must be preceded by `$` as persistent premises do not consume spatial hypotheses, otherwise `processWand` throws a runtime error.                                                                                                                                                                                                          | ✓       |
| Auto-framing for persistent premise (`.autoframe .persistent`)         | For the specialisation pattern `[#$]`.                                                                                                                                                                                                                                                                                                                                                                                                   | ✓       |
| Subgoal for premise with modality (`.goal` with `kind := .modal`)      | For the specialisation pattern `[> $H₁ … $Hₙ ]`. This requires porting the type class `AddModal` and its instances. One of the instances uses `wandM`, which is being ported in PR #470 for `iinv`.                                                                                                                                                                                                                                      | ✓       |
| Auto-framing for premise with modality (`.autoframe .modal`)           | For the specialisation pattern `[>$]`.                                                                                                                                                                                                                                                                                                                                                                                                   | ✓       |

The original `SpecializeState` contains the field `pf : Q($orig ⊢ $e ∗ □?$p $out)`, but `orig ⊢ e ∗ □?p out` does not hold in general when the specialisation patterns involve `.autoframe .modal` or `.goal` with `kind := .modal`. The assertion therefore has to be weakened as `(e ∗ □?p out ⊢ goal) → orig ⊢ goal`. The problem is that `e ∗ □?p out ⊢ goal` is still required by the theorem `specialize_dup_context` for context duplication. The current solution is to have `(e ∗ □?p out ⊢ goal) → orig ⊢ goal` as a mandatory field and `orig ⊢ e ∗ □?p out` as an optional field in `SpecializeState`. Context duplication is possible only when the proof `orig ⊢ e ∗ □?p out` exists. Context duplication is mutually exclusive with specialisation patterns involving modalities.

The missing `AddModal` instances are also being ported. This addresses #237 in full, as well as #255 and #400 in part.

| `AddModal` instance        | Located in ... (Rocq)                                                                                                      | Located in.. (Lean)                                                                                                                | Ported? |
|----------------------------|----------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|---------|
| `add_modal_fupd_wp`        | [`weakestpre.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/program_logic/weakestpre.v)                       | [`WeakestPre.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProgramLogic/WeakestPre.lean)          | ✓       |
| `add_modal_forall`         | [`class_instances.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances.v)                 | [`Instances.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/Instances.lean)               | ✓       |
| `add_modal_tforall`        | `class_instances.v`                                                                                                        | `Instances.lean`                                                                                                                   |         |
| `add_modal_wand`           | `class_instances.v`                                                                                                        | `Instances.lean`                                                                                                                   | ✓       |
| `add_modal_wandM`          | `class_instances.v`                                                                                                        | `Instances.lean`                                                                                                                   | ✓       |
| `add_modal_except_0`       | [`class_instances_later.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_later.v)     | [`InstancesLater.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/InstancesLater.lean)     | ✓       |
| `add_modal_except_0_later` | `class_instances_later.v`                                                                                                  | `InstancesLater.lean`                                                                                                              | ✓       |
| `add_modal_later`          | `class_instances_later.v`                                                                                                  | `InstancesLater.lean`                                                                                                              | ✓       |
| `add_modal_later_except_0` | `class_instances_later.v`                                                                                                  | `InstancesLater.lean`                                                                                                              | ✓       |
| `add_modal_bupd`           | [`class_instances_updates.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_updates.v) | [`InstancesUpdates.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/InstancesUpdates.lean) | ✓       |
| `add_modal_fupd`           | `class_instances_updates.v`                                                                                                | `InstancesUpdates.lean`                                                                                                            | ✓       |
| `add_modal_id`             | [`classes.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/classes.v)                                 | [`Classes.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/Classes.lean)                   | ✓       |

There are also several modules yet to be ported in which `AddModal` instances are located:
- [`program_logic/total_weakestpre.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/program_logic/total_weakestpre.v): `add_modal_fupd_twp`
- [`proofmode/class_instances_embedding.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_embedding.v): `add_modal_embed_bupd_goal`, `add_modal_embed_fupd_goal`
- [`proofmode/monpred.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/monpred.v): `add_modal_at_bupd_goal`, `add_modal_at_fupd_goal`

### Other Changes

- Similar to the changes in PR #496, `SpecPat.parse` is extended to return a pair for `Syntax` and `specPat` values so that `processWand` can use `withRef` to provide more precise syntax highlighting for error messages.

<p align="center">
<img width="204" height="38" alt="More syntax highlighting using withRef" src="https://github.com/user-attachments/assets/7018a86f-70d3-4dbf-a103-d93c0b70f125" />
</p>

- Docstrings are added for the specialisation pattern syntax, similar to the changes in PR #464.
- [Bug fix](https://github.com/leanprover-community/iris-lean/pull/500/commits/3c14a21c560ac903f3af39c269ef97f3698ef299): use `ProofModeM.trySynthInstance` instead of the built-in synthesiser (`trySynthInstance`) for type classes with the `ipm_class` annotation.
- Added missing `ipm_backtrack` annotations to the type class instances `intoAnd_sep`, `intoAnd_sep_affine` and `intoAbsorbingly_absorbing`.
- The function argument `e` for `iFrame` is made implicit as it can always be inferred using `hyps`.

## Checklist


Iris/Iris/ProofMode/Patterns/SpecPattern.lean
```diff
@@ -20,25 +20,61 @@ syntax ident : frameIdent
...
+syntax term (colGt " $$ " (colGt ppSpace specPat)+)? : pmTerm
 
 @[rocq_alias goal_kind]
 inductive SpecGoalKind
   | spatial
-  -- TODO: implement
   | modal
-  -- TODO: implement
-  | intuitionistic
+  | persistent
```
MackieLoeffel: I think this should be called `intuitionistic`

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
```diff
@@ -50,10 +86,9 @@ structure SpecGoal where
   hyps : List Ident
 deriving Repr, Inhabited
 
--- see https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/spec_patterns.v?ref_type=heads#L15
 @[rocq_alias spec_pat]
 inductive SpecPat
-  | ident (name : Ident)
+  | ident (name : Ident) (recursiveSpecPats : List <| Syntax × SpecPat)
```
MackieLoeffel: Can this just be `| ident (t : PMTerm)`?

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
```diff
@@ -50,10 +86,9 @@ structure SpecGoal where
   hyps : List Ident
 deriving Repr, Inhabited
 
--- see https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/spec_patterns.v?ref_type=heads#L15
 @[rocq_alias spec_pat]
 inductive SpecPat
```
MackieLoeffel: Should `SpecPat` contain the `Syntax` and the inductive `SpecPatCase` or so, similar to cases pattern? At the moment there is quite a bit manual threading around of `Syntax` objects.

Iris/Iris/ProofMode/Tactics/Rewrite.lean
```diff
@@ -112,11 +112,16 @@ end rule
 end IRewrite
 
 private def iRewriteCore {prop : Q(Type u)} {bi : Q(BI $prop)}
-    {e} (hyps : Hyps bi e) (rule : IRewrite.Rule)
+    {e} (hyps : Hyps bi e) (_goal : Q($prop)) (rule : IRewrite.Rule)
```
MackieLoeffel: Why is the `_goal` necessary?

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,134 +16,308 @@ namespace Iris.ProofMode
...
 
-structure SpecializeState {prop : Q(Type u)} (bi : Q(BI $prop)) (orig : Q($prop)) where
-  (e : Q($prop)) (hyps : Hyps bi e) (p : Q(Bool)) (out : Q($prop))
-  pf : Q($orig ⊢ $e ∗ □?$p $out)
+structure SpecializeState {prop : Q(Type u)} {bi : Q(BI $prop)} (orig goal : Q($prop)) where
+  {e : Q($prop)} (hyps : Hyps bi e) (p : Q(Bool)) (out : Q($prop))
+  -- A weaker proposition than `pf`, applicable even when the `.modal` kind is involved
+  (pfCont : Q(($e ∗ □?$p $out ⊢ $goal) → $orig ⊢ $goal))
+  -- Holds when the `.modal` kind is not involved, cannot duplicate context if `pf = none`
+  pf : Option Q($orig ⊢ $e ∗ □?$p $out)
```
MackieLoeffel: It is a bit annoying that we have to construct two proofs. Is there a single proof term that works for both? Maybe something like `$e ∗ □?$p $out -∗ $goal ⊢ $orig -∗ $goal`?

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
alvinylt: Renamed in commit [74b6ff4](https://github.com/leanprover-community/iris-lean/pull/500/commits/74b6ff4b5c53405c3c92b88dfeeff81738e4657a).

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
alvinylt: Yes, `SpecPat` and `PMTerm` is now defined by mutual induction.

Commit: [ce55c1c](https://github.com/leanprover-community/iris-lean/pull/500/commits/ce55c1c00ce3f2be55a43fe272ccb971ce8a7ecf)

Iris/Iris/ProofMode/Patterns/SpecPattern.lean
alvinylt: Yes, by `SpecPat`, `PMTerm` and `SpecPatCase` are now defined by mutual induction such that the rest of the code is simplified.

Commit: [832d1d0](https://github.com/leanprover-community/iris-lean/pull/500/commits/832d1d0b0607cd6691b14e88158de26231490688)

Iris/Iris/ProofMode/Tactics/Rewrite.lean
alvinylt: Residual code from previous attempts to update `iRewriteCore`. This redundant function argument has now been removed.

Commit: [ec648cb](https://github.com/leanprover-community/iris-lean/pull/500/commits/ec648cb924e0e4251b025c798e81d3d0eb991705)

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: This one is tricky because `orig ⊢ e ∗ □?p out` is equivalent to `∀ goal, e ∗ □?p out -∗ goal ⊢ orig -∗ goal`, which is strictly stronger than both `(e ∗ □?p out ⊢ goal) → orig ⊢ goal` and `e ∗ □?p out -∗ goal ⊢ orig -∗ goal` whenever `goal` is specified.

In a nutshell, my original approach with two proofs in `SpecializeState` works because the use of modality-related patterns (e.g. `[>$]`) and context duplication are mutually exclusive. To avoid having two proofs in `SpecializeState`, the trick is actually to refactor `iSpecializeCore`:
- We first check whether modality-related specialisation patterns are used, and if so, context duplication is not applicable anyway. Hence, there is no need for the proof of `orig ⊢ e ∗ □?p out`.
- Otherwise, we first create a metavariable `goal` for the proof `(e ∗ □?p out ⊢ goal) → orig ⊢ goal` and then call `processWand`.
  - The metavariable `goal` is eventually instantiated with `e ∗ □?p out ⊢ goal`, so the proof in `SpecializeState` becomes `(e ∗ □?p out ⊢ e ∗ □?p out) → orig ⊢ e ∗ □?p out`. The premise is trivial by reflexivity, so we obtain `orig ⊢ e ∗ □?p out` for context duplication.
  - We rely on the fact that `goal` remains an uninstantiated metavariable throughout the code in `processWand` as long as no modality-related specialisation patterns are involved.

Commit: [b7ff996](https://github.com/leanprover-community/iris-lean/pull/500/commits/b7ff996e55476f3893d0bcb0533bd395dcad3883)

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,133 +16,288 @@ namespace Iris.ProofMode
...
-    let out₁ ← mkFreshExprMVarQ prop
-    let out₂ ← mkFreshExprMVarQ prop
-    let some _ ← ProofModeM.trySynthInstanceQ q(IntoWand $p false $out .out $out₁ .out $out₂)
-      | throwError m!"ispecialize: {out} is not a wand"
-    let res ← iFrame hyps out₁ (← SelPat.resolve hyps [.spatial, .intuitionistic])
+    let pfStep : Q($e ∗ □?$p $out ⊢ $e ∗ □?$p $Φ $x) :=
+      q(specialize_forall (A2 := $e) (p := $p) $inst $x)
+    return specState.update hyps p out' pfStep
+  -- Subgoal with `[ H₁ … Hₙ ]` or `[- H₁ … Hₙ ]`
+  | .goal { kind := .spatial, negate, trivial, frame := f, hyps := hs } g => do
```
MackieLoeffel: The cases for `.spatial`, `.intuitionistic`, and `.modal` are very similar. Can you combine them into one case with some if statements to avoid the duplication? Then one can also maybe avoid some of the helper functions.

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,133 +16,288 @@ namespace Iris.ProofMode
...
+  | .goal { kind := .modal, negate, trivial, frame := f, hyps := hs, .. } g =>
+    let ⟨el, _, hypsl', hypsr', pf', frameIVars⟩ ← splitFrameHyps hyps hs f negate
+    let ⟨_, out2, out1', instWand, instModal⟩ ← synthIntoWandModal p out goal
+    let pf'' ← finishFrameSubgoal hypsr' out1' trivial g frameIVars
+    let h : Q($e ⊢ $el ∗ $out1') := q($(pf').mp.trans (sep_mono_right $pf''))
+    let pfStep : Q(($el ∗ □?false $out2 ⊢ $goal) → $e ∗ □?$p $out ⊢ $goal) :=
+      q(fun k => specialize_modal $h k $instWand $instModal)
+    return specState.updateCont hypsl' q(false) out2 pfStep
+  -- Auto-framing with `[$]`
+  | .autoframe .spatial => do
```
MackieLoeffel: The comment about duplication also applies to the `autoframe` cases. Can you combine them into one case?

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -160,40 +315,75 @@ A tuple containing:
...
+    (spats : List SpecPat) (try_dup_context : Bool := false) :
+    ProofModeM ((e' : _) × Hyps bi e' × (pb : Q(Bool)) × (B : Q($prop)) ×
+      Q(($e' ∗ □?$pb $B ⊢ $goal) → $e ∗ □?$pa $A ⊢ $goal) ×
+      Option Q($e ∗ □?$pa $A ⊢ $e' ∗ □?$pb $B)) := do
+  -- Return the result directly when there are no nested specialisation patterns
+  if spats.isEmpty then
+    return ⟨_, hyps, pa, A, q(id), some q(.rfl)⟩
+
+  -- Modality-related specialisation patterns involved
+  if spats.any (·.anyModal) then
```
MackieLoeffel: We can also use this simple case if `try_dup_context` is false.

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -318,6 +318,42 @@ instance (priority := default - 10) elimModal_timeless [BI PROP] p io (P P' Q :
   ElimModal True p io p P P' Q Q where
   elim_modal _ := ((sep_mono ((intuitionisticallyIf_mono into_except0).trans except0_intuitionisticallyIf) except0_intro).trans $ except0_sep.2.trans (except0_mono wand_elim_right)).trans is_except0
 
+/-- AddModal -/
+@[rocq_alias add_modal_later_except_0]
+instance (priority := default + 10) addModal_later_except_0 [BI PROP] (P Q : PROP) [h : Timeless P] :
```
MackieLoeffel, line 324: This instance probably needs an `ipm_backtrack`

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -318,6 +318,42 @@ instance (priority := default - 10) elimModal_timeless [BI PROP] p io (P P' Q :
...
+    AddModal iprop(▷ P) P iprop(◇ Q) where
+  add_modal := calc
+    _ ⊢ ◇ P ∗ (P -∗ ◇ Q)   := sep_mono_left h.timeless
+    _ ⊢ ◇ (P ∗ (P -∗ ◇ Q)) := except0_frame_right
+    _ ⊢ ◇ (◇ Q)            := except0_mono wand_elim_right
+    _ ⊢ ◇ Q                 := except0_idem.mp
+
+@[rocq_alias add_modal_later]
```
MackieLoeffel: Also this instance likely needs `ipm_backtrack`.

Iris/Iris/ProofMode/Porting.lean
```diff
@@ -26,7 +26,7 @@ import Iris.Std.RocqPorting
 #rocq_concept proofmode "Tactics" "iAssumption" ported "iassumption"
 #rocq_concept proofmode "Tactics" "iAssumptionCoq" ignored "weird tactic"
 #rocq_concept proofmode "Tactics" "iExFalso" ported "iexfalso"
-#rocq_concept proofmode "Tactics" "iPure (basic)" ported "ipure"
+#rocq_concept proofmode "Tactics" "iPure (basic)" ported "ipu<re"
```
MackieLoeffel: Typo?

alvinylt: Yes, fixed in commit [dfce754](https://github.com/leanprover-community/iris-lean/pull/500/commits/dfce7544fc51009b385333e1f05cb53e74d3e966).

Iris/Iris/ProofMode/InstancesLater.lean
alvinylt: Indeed, `ipm_backtrack` is added to the instance. I have also added a new test for `ispecialize` to check that the backtracking works.

Commit: [baca49d](https://github.com/leanprover-community/iris-lean/pull/500/commits/baca49db325e71753e5e2d36deed80bcf91835e4)

Iris/Iris/ProofMode/InstancesLater.lean
alvinylt: Resolved in the same commit above (commit [baca49d](https://github.com/leanprover-community/iris-lean/pull/500/commits/baca49db325e71753e5e2d36deed80bcf91835e4)).

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: This is tricky: the cases `.spatial`, `intuitionistic` and `.modal` are all different in terms of the type classes synthesis steps and theorems used. For any two of the three cases, there is at most one common line of code, but there isn't any part of the code shared by all three cases.

Instead, it makes more sense to deduplicate code with the following pairing:
- `.goal` with `.kind := .spatial` (`[ H1 ... ]`, `[- H1 ... ]`) and `.autoframe` with `.kind := .spatial` (`[$]`)
- `.goal` with `.kind := .intuitionistic` (`[ #H1 ... ]`) and `.autoframe` with `.kind := .intuitionistic` (`[#$]`)
- `.goal` with `.kind := .modal` (`[> H1 ... ]`, `[>- H1 ... ]`) and `.autoframe` with `.kind := .modal` (`[>$]`)

I have refactored the code so that all these cases `processSpecGoal`, which then does pattern matching on `.kind`.

The deduplication is minimal (code reduced by 16 lines), but since we are moving plenty of things out of the mutual induction block, the build time seems to be significantly reduced.

alvinylt: Commit: [0631755](https://github.com/leanprover-community/iris-lean/pull/500/commits/0631755f3f8371d338472bc2a99a3576cdac3043)

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: Addressed in the comment above.

Commit: [0631755](https://github.com/leanprover-community/iris-lean/pull/500/commits/0631755f3f8371d338472bc2a99a3576cdac3043)

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: I initially tried the following directly:

```lean4
 if !try_dup_context || spats.any (·.anyModal) then
    let st ← spats.foldlM processWand { hyps, p := pa, out := A, pf := q(id) }
    return ⟨_, st.hyps, st.p, st.out, st.pf⟩
```

This would break `iinv` as the metavariable trick produces the proof useful not just for context duplication but also for other usage cases such as in `iInvCore`, where `try_dup_context = false`.

The solution to this is some more refactoring so that `iSpecializeCore` calls `iSpecializeCoreNoModal`. The latter function stipulates that there is no modality-related specialisation pattern involved and returns the stronger proof. A side benefit of this is that we are able to simplify the type signature of `iSpecializeCore`. More specifically, `iSpecializeCore` now returns one proof instead of two.

Commit: [fb1b331](https://github.com/leanprover-community/iris-lean/pull/500/commits/fb1b331e489507f6532e6962bbda5c47506d3f19), [c4356fd](https://github.com/leanprover-community/iris-lean/pull/500/commits/c4356fd00c0391a483c015e72fe3f8768b4668a6)

Iris/Iris/ProofMode/Tactics/Specialize.lean
MackieLoeffel: I think you can unify the spatial and modal cases if you use the trivial AddModal instance for the spatial case (i.e. prove `AddModal` for the spatial case using `addModal_id`).

Good observation that moving things out of the mutual induction block reduces build time!

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,133 +16,273 @@ namespace Iris.ProofMode
...
+  let ⟨ref, spat⟩ := spat
+  withRef ref do
+  match spat with
+  -- A hypothesis name, possibly with nested specialisation patterns
+  | .ident pmt =>
+    let ivar ← hyps.findWithInfo ⟨pmt.term⟩
+    let ⟨_, hyps', _, out1', p1, _, pf'⟩ := hyps.remove false ivar
+    let ⟨e'', hyps'', pNest, outNest, pfContNest⟩ ←
+      iSpecializeCore hyps' p1 out1' q(iprop(□?$p $out -∗ $goal)) pmt.spats
+    let p2 := if pNest.constName! == ``true then p else q(false)
```
MackieLoeffel:
```suggestion
    let p2 := if isTrue pNest then p else q(false)
```

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,133 +16,273 @@ namespace Iris.ProofMode
...
+    ProofModeM ((e' : Q($prop)) × Hyps bi e' × Q($e ⊢ $e' ∗ $P)) := do
+  match spec with
+  -- Generate subgoal: `[ H₁ … Hₙ ]`, `[- H₁ … Hₙ ]`, `[> H₁ … Hₙ ]` or `[>- H₁ … Hₙ ]`
+  | some ⟨{ negate, trivial, frame := f, hyps := hs, .. }, name⟩ =>
+    let ⟨el, _, hypsl, hypsr, pf', frameIVars⟩ ← splitFrameHyps hyps hs f negate
+    let pf'' ← finishFrameSubgoal hypsr P trivial name frameIVars
+    return ⟨el, hypsl, q($(pf').mp.trans <| sep_mono_right $pf'')⟩
+  -- Auto-framing: `[$]` or `[>$]`
+  | none =>
+    let res ← iFrame hyps P <| ← SelPat.resolve hyps [.spatial, .intuitionistic]
```
MackieLoeffel: Why does this use `iFrame` instead of `finishFrameSubgoal`? Also `[#$]` seems to use itrivial, but `[$]` and `[>$]` do not as far as I see. I am not fully sure, but I think the Rocq version uses the correspondence of `itrivial` in all three cases.

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -16,133 +16,273 @@ namespace Iris.ProofMode
...
+partial def processWand {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {orig goal : Q($prop)}
+    (specState : @SpecializeState u prop bi orig goal) (spat : SpecPat) :
+    ProofModeM (@SpecializeState u prop bi orig goal) := do
+  let { e, hyps, p, out, .. } := specState
+  let ⟨ref, spat⟩ := spat
+  withRef ref do
+  match spat with
+  -- A hypothesis name, possibly with nested specialisation patterns
+  | .ident pmt =>
+    let ivar ← hyps.findWithInfo ⟨pmt.term⟩
```
MackieLoeffel: What if `pmt.term` is not actually an identifier? (I am not sure if we need / want to support that case, but we should at least make sure that there is a sensible error message. Maybe something similar to the code in `ispecialize`.)

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: Resolved in commit [62984f6](https://github.com/leanprover-community/iris-lean/pull/500/commits/62984f633066bbe0e91ed83de1329655e6a86db2).

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: The line is modified to use `try?` and throws an error if the identifier is invalid.

Commit: [2c16618](https://github.com/leanprover-community/iris-lean/pull/500/commits/2c16618501f8afc85081ba13da2bdb9072ef815f)

Iris/Iris/ProofMode/Tactics/Specialize.lean
alvinylt: My oversight: it's actually the opposite in [the Rocq version](https://gitlab.mpi-sws.org/iris/iris/-/blob/06233e076a9d72f05c41c6059f9fc9e9b98846a0/iris/proofmode/ltac_tactics.v#L999-L1015). With the auto-framing (`[$]`, `[#$]` and `[>$]`) patterns, `itrivial` are not used.

Here is an example proof in Rocq:

```rocq
Lemma test_lemma (P Q : PROP) : (⌜2 = 2⌝ -∗ P) -∗ P.
Proof.
  iIntros "Himp".
  iSpecialize ("Himp" with "[# //]").
  iAssumption.
Qed.
```

It fails with `iSpecialize ("Himp" with "[# $]")`.

My previous implementation of `[#$]` was too strong and solved it. Now it no longer does (consistent with the other two auto-framing patterns).

Commit: [0e02267](https://github.com/leanprover-community/iris-lean/pull/500/commits/0e0226773dc19c5fd27a01c046c72593f8a121de)

Iris/Iris/ProofMode/Tactics/Specialize.lean
MackieLoeffel: I cleaned this up in https://github.com/leanprover-community/iris-lean/pull/500/commits/0d24861beca233f24f35138ccbf3ac32e89c2614

MackieLoeffel:
I made some small cleanups in https://github.com/leanprover-community/iris-lean/pull/500/commits/0d24861beca233f24f35138ccbf3ac32e89c2614. It looks good from my side now! @alvinylt Is there anything you want to do or can this be merged?

alvinylt:
Looks good to me! This can be merged.

MackieLoeffel:
Great! Thanks for your work on this, I think this is really cool! Merging.

PR #502, opened 2026-07-06
feat: rm setoids (step 1)

Kaptch:
## Description
- Changes definition of OFE to relate distance to equality, instead of equivalence, with that, all OFE become Leibniz. The old definition of Equiv is replaced by \forall n, Dist n x y.
- Removes Leibniz and LeibnizPreserving (not needed anymore).
- Adds to_eq to BI interface, now it's not a property of particular instances, but part of the interface. 
- Fixes downstream proofs (mostly in the Algebra folder). 
- Short demo is in BI/Algebra, where a lot of annoying manual Equiv constructions were factored out. 

This is the first step, next, I'm going to start from the leaves of the project replacing Equiv with equality, manual Eequiv constructions with simp/rw tactics to avoid big PRs. 
Apart from the listed changes, I tried to avoid changing any interfaces, so it can be used by downstream projects only with the proposed changes to the OFE interface, and minor changes to Equiv-related proofs (if it was unfolding definition of a particular Equiv).

Also, I've not included it in the PR, but now if you import Mathlib, it becomes easy to show that (C)OFEs are a subclass (complete) ultrametric spaces, and it hopefully can be used for reusing some topology results from Mathlib (I can include it later, but probably it's better if this PR stays relatively small).

### DIFF
--- a/Iris/Iris/Algebra/Agree.lean
+++ b/Iris/Iris/Algebra/Agree.lean
@@ -110,25 +110,15 @@ theorem dist_equiv : Equivalence (dist (α := α) n) where
       obtain ⟨a, ha, hd₂⟩ := h₁' b hb
       exact ⟨a, ha, hd₂.trans hd₁⟩
 
-instance instOFE : OFE (Raw α) where
-  Equiv x y := ∀ n, dist n x y
-  Dist := dist
-  dist_eqv := dist_equiv
-  equiv_dist := by simp
-  dist_lt {n x y m} := fun ⟨h₁, h₂⟩ hlt => by
-    refine ⟨?_, ?_⟩
-    · intro a ha
-      obtain ⟨b, hb, hd⟩ := h₁ a ha
-      exact ⟨b, hb, OFE.Dist.lt hd hlt⟩
-    · intro b hb
-      obtain ⟨a, ha, hd⟩ := h₂ b hb
-      exact ⟨a, ha, OFE.Dist.lt hd hlt⟩
-
-theorem dist_lt {x y : Raw α} (h : dist n x y) (hlt : m < n) : dist m x y :=
-  instOFE.dist_lt h hlt
-
-theorem equiv_def {x y : Raw α} : x ≡ y ↔ ∀ n, dist n x y := .rfl
-theorem dist_def {x y : Raw α} : x ≡{n}≡ y ↔ dist n x y := .rfl
+theorem dist_lt {x y : Raw α} (h : dist n x y) (hlt : m < n) : dist m x y := by
+  obtain ⟨h₁, h₂⟩ := h
+  refine ⟨?_, ?_⟩
+  · intro a ha
+    obtain ⟨b, hb, hd⟩ := h₁ a ha
+    exact ⟨b, hb, OFE.Dist.lt hd hlt⟩
+  · intro b hb
+    obtain ⟨a, ha, hd⟩ := h₂ b hb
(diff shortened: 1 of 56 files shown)


Kaptch:
Oops, forgot that we also have IrisMath. @markusdemedeiros could you take a look at the random variable example? I think it essentially the same, as it was before (modulo quotienting before providing OFE instance), but I am not sure how it is used downstream.

markusdemedeiros:
Taking a look at this this afternoon!

Iris/Iris/Algebra/DFrac.lean
```diff
@@ -80,36 +79,16 @@ instance instCMRADFrac : CMRA DFrac where
...
-        exact ⟨.rfl, .rfl, .rfl⟩
-      · exists own y, ownDiscard z
-      · exists discard, own x
-        obtain rfl : x = z := Subtype.ext Hxyz'
-        exact ⟨.rfl, .rfl, .rfl⟩
-      · exists discard, ownDiscard x
-      · exists ownDiscard y, own z
-      · exists ownDiscard x, discard
-      · exists ownDiscard y, ownDiscard z
+  extend _ Hxyz := ⟨_, _, discrete Hxyz, .rfl, .rfl⟩
```
markusdemedeiros, line 91: Lovely :)

Iris/Iris/Algebra/Agree.lean
```diff
@@ -238,18 +229,17 @@ theorem toAgree_uninj {x : Raw α} : x.valid → ∃ a, ∀ n, dist n (toAgree a
...
-    refine ⟨?_, ?_⟩
-    · simp only [map', List.mem_map, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂]
-      intro a ha
-      obtain ⟨b, hb, heq⟩ := h.1 a ha
-      exact ⟨f b, ⟨b, hb, rfl⟩, OFE.NonExpansive.ne heq⟩
-    · simp only [map', List.mem_map, forall_exists_index, and_imp, forall_apply_eq_imp_iff₂]
-      intro a ha
-      obtain ⟨b, hb, heq⟩ := h.2 a ha
-      exact ⟨f b, ⟨b, hb, rfl⟩, OFE.NonExpansive.ne heq⟩
+theorem map'_ne [OFE.NonExpansive f] {x₁ x₂ : Raw α} (h : dist n x₁ x₂) :
```
markusdemedeiros, line 232: Why is this not a NonExpansive instance anymore? Or at least stated using NonExpansive?

markusdemedeiros (approved):

markusdemedeiros:
Superb. I had one question, but only to check if one change is intentional or not. Thanks for making this change in its own PR!

Iris/Iris/Algebra/Agree.lean
Kaptch: Agree.Raw is not an OFE anymore (it could with the old OFE definition, but here pre-quotient it can't), only the quotiented version `Agree` is, so we can't use normal `NonExpansive` here. I just didn't come up with a better name for this lemma :)

Kaptch:
Just in case, could we wait with merging it until Friday (just for one final discussion, since it's a relatively big divergence with the Rocq version)? I'll update it to the current master, so no worries about that.

markusdemedeiros:
Sure, sounds good with me.

lzy0505 (approved):

PR #506, opened 2026-07-09
Fix: `simp` and `rw` sometimes exit the ProofMode

### LINKED ISSUE #469: `simp` can exit the ProofMode
There are some examples in our repo where `simp` will unfold too far, exiting the Iris proofmode altogether. At the moment the fix is is to use `dsimp` instead, but the proofmode should probably be stable under `simp`. The bug can be witnessed by changing the `dsimp only` statements to `simp only` in the (post 4.31) Quicksort example, though I recall encountering this it in a few places even before the 4.31 upgrade. 

Michael suggests making `Entails'` a `def` instead of an `abbrev`, which seems like a good idea, but currently breaks the build in a number of places.

lzy0505:
## Description

Fixes #469. The hack seems to work.

As I wrote in the comments, I think it is `kabstract` that causes the issue. It is probably a bug?

## Checklist

## Generative AI Guidelines


markusdemedeiros:
I think I'm not the right one to evaluate this, but it seems reasonable?

MackieLoeffel:
Thanks, this fix looks good to me. I slightly cleaned up the PR. From my side, this can be merged.

Issue #239, opened 2026-04-07
Port proofmode/coq_tactics.v and ltac_tactics.v


lzy0505:
## Summary

Port the remaining items from [`proofmode/coq_tactics.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/coq_tactics.v) and [`proofmode/ltac_tactics.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/ltac_tactics.v).

> **Note:** Split into the files in `Tactics/` in the Lean port.

## Subtasks

- [ ] iStartProof with bi specified
- [ ] iEval
- [ ] iSimpl
- [ ] iUnfold
- [ ] iPure: pure destructuring patterns (also for other tactics using ipure)
- [ ] iFrame #358 
- [ ] iRevert: `intoIH` revert
- [ ] iSpecialize: all specialization patterns
- [ ] iNext (with later credits)
- [ ] iDestruct (icases): all destruct patterns
- [ ] iCombine #359 
- [ ] iIntros: all intro patterns
- [ ] iInduction #360 
- [ ] iLöb #361
- [ ] iRewrite #362
- [ ] iInv #363
- [ ] iAccu #364
- [ ] Rules for trivial

## Already ported

- [x] iSolveSideCondition
- [x] iStartProof (basic)
- [x] iStopProof
- [x] iRename
- [x] iClear (basic + selection patterns)
- [x] iExact
- [x] iAssumption
- [x] iExFalso
- [x] iPure (basic)
- [x] iEmpIntro
- [x] iPureIntro
- [x] iRevert (basic + selection patterns)
- [x] iPoseProof (Lean: ihave _ := _)
- [x] iSpecialize (basic + duplicate context for persistent conclusion)
- [x] iApply
- [x] iLeft
- [x] iRight
- [x] iSplit(L/R)
- [x] iExists
- [x] iModIntro
- [x] iMod
- [x] iDestruct (basic)
- [x] iIntros (basic)
- [x] iAssert (Lean: ihave _ : _)

## Dependencies

**Rocq dependencies:**

- `bi/bi` (aggregation)
- #211
- #219
- #212
- #213
- #225
- #215

PR #510, opened 2026-07-10
feat: inext with later credits

alvinylt:
## Description

Implements the tactic `inext n credit: H`.

Addresses #239.

Similar to the Rocq implementation, once all credits are consumed (that is, `H` becomes `£ 0`), the hypothesis is discarded.

~~The type class `AddModal` and its instances have already been ported in PR #470, but are nonetheless also included in this PR so that we can use them for this tactic.~~ _Using `ElimModal` instead._

~~This tactic is unique in the sense that it is relevant only for `IProp GF` where `InvGS GF` holds. Within the `ProofModeM.runTactic` block, we have the generic BI instance `prop`. Otherwise, if we use `isDefEq` to unify `prop` and `InvGS GF` manually, we still run into universal level problems:~~
```
Application type mismatch: The argument
  «$GF»
has type
  BundledGFunctors.{u✝, u✝, u✝}
of sort `Type (u✝ + 1)` but is expected to have type
  BundledGFunctors.{0, 0, 0}
of sort `Type 1` in the application ...
```

~~In the implementation, we use `mkAppM` so that the instance `InvGS GF` is synthesised automatically.~~ _Resolved in commits [9056733](https://github.com/leanprover-community/iris-lean/pull/510/commits/905673382ef322b2a792f4dc594c5f2ffecd501a) and [9b1963a](https://github.com/leanprover-community/iris-lean/pull/510/commits/9b1963abf69586da8892d763f358a6e00b7d30c4)._

## Checklist


Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,109 @@ end StepIndexed
...
+          | lc GF _  _ c =>
+            match c.nat? with
+            | none => throwError "inext: {c} is not a numeral"
+            | some natM => return some (name, GF, natM)
+          | _ => return none
+    | throwError m!"inext: {h} is not a spatial later credit hypothesis"
+
+    -- To avoid non-termination caused by type class instance check in subsequent steps
+    let some _ ← mkAppM ``InvGS #[GF] >>= (synthInstance? ·)
+    | throwError "inext: requires an InvGS (HasLC) context"
```
MackieLoeffel: This can be simplified to something like the following. Note that this does not give the `name`. Maybe the name should be added to `RemoveHyp`?
```suggestion
    let ⟨e', hyps', out, out', p, _, pfEq⟩ := hyps.remove false ivar
    if isTrue p then throwError "inext: {h} is not in the spatial context"
    let_expr lc GF _  _ c := out' | throwError m!"inext: {h} is not a spatial later credit hypothesis"
    let some natM := c.nat? | throwError "inext: {c} is not a numeral"

    -- To avoid non-termination caused by type class instance check in subsequent steps
    let some _ ← synthInstance? (← mkAppM ``InvGS #[GF])
    | throwError "inext: requires an InvGS (HasLC) context"
```
(Ideally, this would use some form of `Hyps.replace` such that the later credits hypothesis keeps its spot in the context. But I am not sure how to best do this.)

alvinylt: This part has been simplified. The shifting of hypothesis order in the context also occurs with some other tactics, so I think this is a broader issue to be addressed by improving the structure and management of `Hyps`.

Commit: [4240629](https://github.com/leanprover-community/iris-lean/pull/510/commits/4240629f409e9e32df8386787e8c167930ad2856)

alvinylt:
The latest improvement is to use [the same trick for HeapLang tactics](https://github.com/leanprover-community/iris-lean/blob/6a9d0cf/Iris/Iris/HeapLang/ProofMode.lean#L182-L187) for matching `IProp GF` with `prop`, so that we can use Qq for building proofs in subsequent steps instead of inelegantly using `mkApp`.

Commits: [9056733](https://github.com/leanprover-community/iris-lean/pull/510/commits/905673382ef322b2a792f4dc594c5f2ffecd501a) and [9b1963a](https://github.com/leanprover-community/iris-lean/pull/510/commits/9b1963abf69586da8892d763f358a6e00b7d30c4)

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,116 @@ end StepIndexed
...
 
+public section
+
+open Lean Elab Tactic Meta Qq Iris.BI Iris Iris.ProofMode
+
+@[rocq_alias tac_lc_add_laterN_split]
+theorem tac_lc_add_laterN_split {GF : BundledGFunctors} [InvGS GF]
+    {n m newM : Nat} {E : CoPset} {P Q goal : IProp GF}
+    (inst : AddModal iprop(|={E}=> goal) goal goal)
+    (h1 : m = n + newM) (h2 : iprop(P ∗ £ newM) ⊢ ▷^[n] Q) (h3 : Q ⊢ goal) :
```
MackieLoeffel: `iprop` is unnecessary here and below

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,116 @@ end StepIndexed
...
+    let ~q(£ $c) := ty
+      | throwError m!"inext: {h} is not a spatial later credit hypothesis"
+    let some natM := c.nat? | throwError "inext: {c} is not a numeral"
+    let ⟨e', hyps', _, _, _, _, pfEq⟩ := hyps.remove false ivar
+
+    -- Ensure sufficient credits
+    if natM < natN then throwError "inext: insufficient credits"
+
+    let g ← mkFreshExprMVarQ q($prop)
+    let some inst ← ProofModeM.trySynthInstanceQ q(AddModal $g $goal $goal)
```
MackieLoeffel: I am a bit confused that this (and also the Rocq code) uses `AddModal` here. It would seem that `ElimModal` would be more fitting since we want to eliminate a fupd modality. Is it possible to change this code to use `ElimModal` instead of `AddModal`?

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,116 @@ end StepIndexed
...
+      let ⟨_, newHyps', pfModAction⟩ ← iModAction hyps' modality
+      let pf ← addBIGoal newHyps' goal
+      have pfEq : Q($e ⊣⊢ $e' ∗ £ $n) := pfEq
+      let pf' : Q($e' ∗ £ $n ⊢ $goal) := q(tac_lc_add_laterN_full $inst $pfModAction $pf)
+      mvar.assign q($(pfEq).mp.trans $pf')
+    -- Update the later credits hypothesis and introduce it into the context
+    | _ =>
+      let newM : Q(Credit) ← pure <| mkNatLit (natM - natN)
+      let newTy : Q($prop) := q(£ $newM)
+      let newHyps := Hyps.add _ name ivar q(false) newTy hyps'
```
MackieLoeffel: This currently does not preserve the order of hypotheses in the context. I don't think we currently have good infrastructure to fix this, so this is fine for this PR. But it would be nice to fix this in the future (across all tactics that replace hypotheses). Maybe we can have a remove function that replaces the hypothesis with emp instead of removing it and then one can later place the updated hypothesis back?

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: Removed in commit [d2a4dc1](https://github.com/leanprover-community/iris-lean/pull/510/commits/d2a4dc11bc19fe41499d6371c82e8eb021a58936).

(I assume you're referring to just `iprop(P ∗ £ newM)`. The `iprop(...)` wrapper is still necessary for `|={E}=> goal`.)

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: I'm not sure about the benefit of switching type class: the proposition `Q'` is an output parameter in `ElimModal`, whereas in our case the proof goal is known and remains unchanged throughout the tactic. Also, the value `E : CoPset` in `iprop(|={E}=> goal)` is unknown, which is why we have `g` created as the metavariable (matching the first argument for `AddModal` as the output parameter):

```lean4
    let g ← mkFreshExprMVarQ q($prop)
    let some inst ← ProofModeM.trySynthInstanceQ q(AddModal $g $goal $goal)
```

On the contrary, `P` is an input parameter of `ElimModal`, so I'm unsure how it can be implemented with `ElimModal`. Do you have any other ideas?

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: This is something I've been aware of from the beginning: `icombine` removes two hypotheses and adds one. In other cases hypotheses also shuffle around. Perhaps something to investigate if we are considering a improvement/redesign of `Hyps`.

Iris/Iris/Instances/Lib/FUpd.lean
MackieLoeffel: Actually at the moment, inputs in IPM TC search cannot be a top-level meta-variables but they can contain meta variables further down and these will be instantiated. So I think something like `[ElimModal ... (|={?E}=> ...) ?P goal ...]` could work (maybe with `.in` as in `elimInv_acc_with_close`.

alvinylt: That makes sense. I have replaced the use of `AddModal` with `ElimModal`, with `E : Q(CoPset)` being a metavariable created before the synthesis.

`ElimModal` comes with a pure side condition `φ`, which is handled by the tactic using `iSolveSidecondition`. With this tactic, the applicable instances are e.g. `elimModal_fupd_fupd` and `elimModalFupdWp`, where the side condition is trivial (`True`).

While `Q'` is an output parameter, the proof goal should remain the same. Hence, we expect `Q'` and `goal` to be definitionally equal (`isDefEq Q' goal do throwError ...`).

Commit: [23dd74d](https://github.com/leanprover-community/iris-lean/pull/510/commits/23dd74d11c0dd446c162292712ba26369443486f)

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -650,21 +649,16 @@ elab "inext" n:(ppSpace num)? " credit: " h:ident : tactic => do
...
+    let E ← mkFreshExprMVarQ q(CoPset)
+    let Q' ← mkFreshExprMVarQ q($prop)
+    let some inst ← ProofModeM.trySynthInstanceQ q(ElimModal $φ false .in false iprop(|={$E}=> $goal) $goal $goal $Q')
+    | throwError "inext: ElimModal type class synthesis failed with {goal}"
+    unless ← isDefEq Q' goal do
+      throwError "inext: eliminating the fancy update does not preserve the goal {goal}"
+    have inst : Q(ElimModal $φ false .in false iprop(|={$E}=> $goal) $goal $goal $goal) := inst
 
-    have inst : Q(AddModal iprop(|={$E}=> $goal) $goal $goal) := inst
+    let hφ ← iSolveSidecondition q($φ) (failOnUnsolved := false)
```
MackieLoeffel: You can remove `(failOnUnsolved := false)`

alvinylt: Resolved in commit [680185f](https://github.com/leanprover-community/iris-lean/pull/510/commits/680185f8de3ad83c418bbfc3f1f6921402bb0621).

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,110 @@ end StepIndexed
...
 
+public section
+
+open Lean Elab Tactic Meta Qq Iris.BI Iris Iris.ProofMode
+
+@[rocq_alias tac_lc_add_laterN_split]
+theorem tac_lc_add_laterN_split {GF : BundledGFunctors} [InvGS GF]
+    {φ : Prop} {n m newM : Nat} {E : CoPset} {P Q goal : IProp GF}
+    (inst : ElimModal φ false .in false iprop(|={E}=> goal) goal goal goal) (hφ : φ)
+    (h1 : m = n + newM) (h2 : iprop(P ∗ £ newM) ⊢ ▷^[n] Q) (h3 : Q ⊢ goal) :
```
MackieLoeffel:
```suggestion
    (h1 : m = n + newM) (h2 : P ∗ £ newM ⊢ ▷^[n] Q) (h3 : Q ⊢ goal) :
```

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,110 @@ end StepIndexed
...
+    inext
+    imodintro
+    iapply h3 $$ H
+  · iintro _ //
+
+theorem tac_lc_add_laterN_full {GF : BundledGFunctors} [InvGS GF]
+    {φ : Prop} {m : Nat} {E : CoPset} {P Q goal : IProp GF}
+    (inst : ElimModal φ false .in false iprop(|={E}=> goal) goal goal goal) (hφ : φ)
+    (h2 : P ⊢ ▷^[m] Q) (h3 : Q ⊢ goal) :
+    iprop(P ∗ £ m) ⊢ goal := by
```
MackieLoeffel:
```suggestion
    P ∗ £ m ⊢ goal := by
```

Iris/Iris/ProgramLogic/WeakestPre.lean
```diff
@@ -680,6 +680,13 @@ instance elimModalFupdWpAtomic_wrongMask :
     p io false iprop(|={E₁,E₂}=> P) iprop(False) (WP e @ s ; E₁ {{ Φ }}) iprop(False) where
   elim_modal := nofun
 
+@[rocq_alias add_modal_fupd_wp]
+instance addModalFupdWp : AddModal iprop(|={E}=> P) P (WP e @ s ; E {{ Φ }}) where
```
MackieLoeffel, line 687: Please remove all `AddModal` stuff from this MR now that it is no longer needed here.

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +583,110 @@ end StepIndexed
...
+      | throwError "Expected the BI implementation of `IProp` to be `UPred.instBIUPred`"
+
+    -- Search for the later credit hypothesis from the context
+    let ivar ← hyps.findWithInfo h
+    let some ⟨name, _, p, ty⟩ := hyps.getDecl? ivar
+      | throwError m!"inext: unknown hypothesis {h}"
+    if isTrue p then throwError "inext: {h} is not in the spatial context"
+    let ~q(£ $c) := ty
+      | throwError m!"inext: {h} is not a spatial later credit hypothesis"
+    let some natM := c.nat? | throwError "inext: {c} is not a numeral"
```
MackieLoeffel: Instead of working just for concrete natural numbers, this should ideally use the `NatCancel` typeclass like in Rocq. How hard would it be to adapt this tactic to use `NatCancel`?

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: Unintentionally brought `iprop(...)` back during the recent switch from `AddModal` to `ElimModal`...

Resolved in commit [abab86c](https://github.com/leanprover-community/iris-lean/pull/510/commits/abab86ca52fd95a5688940e893f5b985880d5a94).

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: Resolved in commit [abab86c](https://github.com/leanprover-community/iris-lean/pull/510/commits/abab86ca52fd95a5688940e893f5b985880d5a94).

Iris/Iris/ProgramLogic/WeakestPre.lean
alvinylt: Resolved in commit [451ddb8](https://github.com/leanprover-community/iris-lean/pull/510/commits/451ddb88ebcfc4b6b2877e2858a206020534a2cb). The PR is now much smaller.

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: I have modified it to use `NatCancel` (was not aware that it exists in `Std.TC`), which simplifies the code as we don't have to use `mkNatLit`, `mkDecideProof`, etc. directly.

Commit: [8011adc](https://github.com/leanprover-community/iris-lean/pull/510/commits/8011adc584c8fe4c522b12767ca52db19385408d)

Iris/Iris/Instances/Lib/FUpd.lean
```diff
@@ -583,4 +584,116 @@ end StepIndexed
...
+    iapply lc_fupd_add_laterN m $$ Hcred
+    inext
+    imodintro
+    iapply h3 $$ H
+  · iintro _ //
+
+public meta section
+
+elab "inext" n:(ppSpace num)? " credit: " h:ident : tactic => do
+  let natN := match n with | none => 1 | some n => n.raw.toNat
```
MackieLoeffel: `inext` should also work when the argument is not a literal number. If you have `H : £ (n + m)`, you should be able to do `inext n credit:H`.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2334,6 +2334,70 @@ example [BI PROP] (P : PROP) : P ⊢ P := by
...
+#guard_msgs in
+example (E : CoPset) (P : IProp GF) : ⊢ £ 1 -∗ ▷ (|={E}=> P) -∗ |={E}=> P := by
+  iintro Hcred HP
+  inext 2 credit: Hcred
+
+/- Tests `inext` with multiple credits consumed at once. -/
+example (E : CoPset) (P : IProp GF) : ⊢ £ 3 -∗ ▷▷▷ (|={E}=> P) -∗ |={E}=> P := by
+  iintro Hcred HP
+  inext 3 credit: Hcred
+  iassumption
```
MackieLoeffel, line 2747: Please add a test that partially uses a concrete credit number. E.g. `£ 3` and then `inext 1` twice or three times.

alvinylt: Test added in commit [5780d22](https://github.com/leanprover-community/iris-lean/pull/510/commits/5780d22a002574932cd8a60e6f3bc20f11747c32).

Iris/Iris/Instances/Lib/FUpd.lean
alvinylt: Support for this feature requires more effort than I expected.

The change within the tactic implementation is simple. We allow the tactic argument to be an arbitrary term and replace the line `let natN := ...` with code mapping the term to a natural number. This already allows simple tests to pass:

```lean4
/- Tests `inext` with multiple credits consumed. -/
example (E : CoPset) (P : IProp GF) :
    ⊢ £ (m + n + 6) -∗ ▷^[m + n + 6] (|={E}=> P) -∗ |={E}=> P := by
  iintro Hcred HP
  inext 3 credit: Hcred
  inext (1 + (3 - 2)) credit: Hcred
  inext 1 credit: Hcred
  inext n credit: Hcred
  inext m credit: Hcred
  iassumption
```

However, `NatCancel` in `Std/TC.lean` is not on the level of generality as that in stdpp. Here is an example where the tactic fails:

```lean4
/- Tests `inext` for later credits with later modalities expressed in terms of `Nat` variables. -/
example (m n p q : Nat) (E : CoPset) (P : IProp GF) :
    ⊢ £ (1 + m + n + p + q + 3) -∗ ▷^[n + m + 4 + p + q] (|={E}=> P) -∗ |={E}=> P := by
  iintro Hcred HP
  inext (m + q) credit: Hcred
  inext (p + n) credit: Hcred
  inext 4 credit: Hcred
  iassumption
```

So what I did was to port `MakeNatAdd`, `MakeNatS`, `NatCancelL` and `NatCancelR`, as well as their instances. The tricky part is to get the priorities of instances right.

With all these classes and instances implemented, the test above passes. One subtle issue was that arithmetic expressions are sometimes not simplified. As a result, we get (correct, but inelegant) expressions such as:

```lean4
HP : ▷^[n + (1 + 1 + 1 + 1) + p]|={E}=> P
```

The solution is therefore to apply `Lean.Meta.evalNat` within `iModAction`. I have added a helper function `normNatArith` for that.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -2334,6 +2334,99 @@ example [BI PROP] (P : PROP) : P ⊢ P := by
...
+/- Tests `inext` for later credits with later modalities expressed in terms of `Nat` variables. -/
+example (m n p q : Nat) (E : CoPset) (P : IProp GF) :
+    ⊢ £ (1 + m + n + p + q + 3) -∗ ▷^[n + m + 4 + p + q] (|={E}=> P) -∗ |={E}=> P := by
+  iintro Hcred HP
+  inext (m + q) credit: Hcred
+  inext (p + n) credit: Hcred
+  inext 4 credit: Hcred
+  iassumption
+
+/- TODO: in the Rocq version, `intoLaterN_later` fails to apply, and `intoLaterN_laterN_bool` applies instead -/
```
alvinylt: This is a test that is supposed to use `intoLaterN_laterN_bool`. This instance is not yet ported, so I have done so along with the addition of this test.

As mentioned in a comment in the Rocq version, it might cancel a later "too much" in the worst case, so this instance has a lower priority than `intoLaterN_later` and `intoLaterN_laterN`.

In the Rocq version, neither `into_laterN_later` nor `into_laterN_laterN` applies to this test, so `into_laterN_laterN_bool` applies, and we get `H : P`. In contrast, `BIBase.laterIf` is a reducible `def` in Lean, so `▷?p P` is equivalent to `laterN (Bool.toNat p) P`. As a result, `into_laterN_later` applies (given it has a higher priority than `into_laterN_laterN_bool`).

I'm not sure whether this is intentional or a bug: `MaybeIntoLaterN` is [explicitly ignored](https://github.com/leanprover-community/iris-lean/blob/6acb3bf/Iris/Iris/ProofMode/Classes.lean#L231-L233) and, unlike the Rocq version, `intoLaterN_later` and `intoLaterN_laterN` does not check progress using `TCIf` and `TCEq`. Is there an intended workaround for this, or should we have ported `MaybeIntoLaterN` and implemented the instances faithfully?

alvinylt: In commit [0b7d5fb](https://github.com/leanprover-community/iris-lean/pull/510/commits/0b7d5fb9c3a69bac119020ec73dec9d26f8cc5be), I tried to implement the design in Rocq faithfully with `TCIf` introduced into `Std/TC.lean` and `MaybeIntoLaterN` being a type class. Basically, `IntoLaterN` is the type class that necessitates progress, whereas `MaybeIntoLaterN` allows a fallback option that enables propositions to remain unchanged. This is intended to avoid the lack of progress being silently accepted without trying other instances. In the example test above, the intended behaviour should be:
1. The instance `intoLaterN_later` has a higher priority and tried first.
2. No progress is made with `intoLaterN_later`, instead of leaving the `H` unchanged, backtrack and try other instances.
3. The instance `intoLaterN_laterN_bool` is tried. Progress is made, so the result of the synthesis is adopted.

This approach with the use of `TCIf`, however, broke the tactic as it is not an IPM type class (because it is meant to be in `Std/TC.lean`). With reference to the type class synthesis trace, the built-in type class synthesiser in Lean tries the applicable instances of `IntoLaterN`. If they all fail, the fallback option with `MaybeIntoLaterN` applies, and the `TCIf ...` instance overall succeeds. And when this propagates to the recursive type class search caller, it uses this default result of unchanged proposition directly. The search could possibly have backtracked such the recursive search had a non-default result.

In fact, there seems to be some trick for `TCIf` in Rocq's stdpp to avoid this:

```rocq
Global Hint Extern 0 (TCIf ⇒
  first [notypeclasses refine (TCIf_true _ _ _ _ _); [tc_solve|]
        |notypeclasses refine (TCIf_false _ _ _ _)] : typeclass_instances.
```

So the solution is similar: instead of using `TCIf`, we define a wrapper type class `GuardedIntoLaterN` which is an IPM type class. Its two instances (`guardedIntoLaterN_stuck` and `guardedIntoLaterN_progress`) are _not_ labelled with `ipm_backtrack`. Meanwhile, instances of `IntoLaterN` and `MaybeIntoLaterN` are labelled with `ipm_backtrack` and adjusted accordingly.

alvinylt:
Now trying to implement `IntoLaterN` and `NatCancel` using `ipm_tactic_instance`.

alvinylt:
The type classes `IntoLaterN`, `MaybeIntoLaterN` and `GuardedIntoLaterN` are consolidated into a single type class. A new parameter (`strict`) for `IntoLaterN` indicates whether we would like progress to made.

The tactic instance `intoLaterNLater` handles the work of `intoLaterN_later`, `intoLaterN_laterN` and `intoLaterN_laterN_bool`, using `isDefEq` to check whether progress is made.

See diff [here](https://github.com/leanprover-community/iris-lean/pull/510/changes/399c12e0a389d3f6e312b99e10a904be073dbb9a..461b5ce88438cadae930bafab1103637f5c08734).

alvinylt:
`NatCancel` has now also been refactored.
- The helper type classes `MakeNatAdd` and `MakeNatS` are no longer necessary and thus removed. Instead we now have two functions to parse expressions (`tryParseSucc` and `tryParseAdd`) and break them into subexpressions.
- `NatCancel` is now an IPM type class. Instead of 25 type class instances, there is a single type class instance (`ipm_class_instance`). Nonetheless there are helper functions to help with the recursion.
- `NatCancel` is from stdpp, so it was originally ported in `Std/TC.lean`. Since it now references features in the IPM type class synthesiser, the code is now under the `ProofMode` directory avoid circular dependency between `ProofMode` and `Std`.

To be honest this new approach isn't really that simpler compared to the original port, especially with all the wrestling with Qq proofs, but is (hopefully) nevertheless easier to understand.

Commit: [4010144](https://github.com/leanprover-community/iris-lean/pull/510/commits/40101446d9239e1ccfa3587d2d627859f3852854)

Iris/Iris/ProofMode/Tactics/ModIntro.lean
```diff
@@ -78,6 +78,22 @@ private def parseModalityActionQ {prop1 prop2 : Q(Type u)} (act : Q(ModalityActi
   | ModalityAction.id _ => return .id
   | _ => throwError "imodintro: unknown modality action {act}"
 
+/--
+  For recursively simplifying arithmetic expressions (e.g. from `m + 1 + 1 + 1`
+  to `m + 3`. We cannot use `whnf` because it rewrites `HAdd.hAdd` into `Nat.add`,
+  which breaks `inext` for later credits.
+-/
+private def normNatArith (e : Expr) : MetaM Expr := transform e (post := postFn)
```
MackieLoeffel: This should not be part of imodintro, but part of the transformation.

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -356,81 +356,188 @@ instance addModal_except_0_later [BI PROP] (P Q : PROP) :
...
+    (h2 : IntoLaterN strict' only_head n' P Q)
+    (h3 : MakeLaterN m' Q lQ) : IntoLaterN strict only_head n iprop(▷^[m] P) lQ where
+  into_laterN := calc
+      _ ⊢ ▷^[m] ▷^[n']Q := laterN_mono _ h2.into_laterN
+      _ ⊢ ▷^[m + n']Q    := (laterN_add _ _).mpr
+      _ ⊢ ▷^[n] ▷^[m']Q := by rw [Nat.add_comm, h1.nat_cancel]; exact (laterN_add _ _).mp
+      _ ⊢ ▷^[n]lQ        := laterN_mono _ h3.make_laterN.mp
+
+@[ipm_backtrack, rocq_alias into_laterN_laterN_bool]
+theorem intoLaterN_laterN_bool [BI PROP] strict {strict'} only_head n (p : Bool) n' m' (P Q lQ : PROP)
```
MackieLoeffel: Try out if removing the reducible annotation on `laterIf` can simplify this.

Iris/Iris/ProofMode/NatCancel.lean
```diff
@@ -0,0 +1,158 @@
...
+  if kb? == some 0 then return (a, q((Nat.add_zero $a).symm))
+  if let ⟨some ka, some kb⟩ := (ka?, kb?) then
+    let s : Q(Nat) := mkNatLit (ka + kb)
+    return (s, ← mkEqRefl s)
+  else
+    let s : Q(Nat) := q($a + $b)
+    return (s, ← mkEqRefl s)
+
+/-- Given `e`, return `e'` such that `e = e' + 1`, if possible. -/
+def tryParseSucc (e : Q(Nat)) : MetaM (Option Q(Nat)) := do
```
MackieLoeffel: Get rid of the succ case and instead parse `n.succ` as a `n + 1` and handle this in the add case.

Iris/Iris/ProofMode/NatCancel.lean
```diff
@@ -0,0 +1,158 @@
...
+
+/-- Given `e`, return `e1` and `e2` such that `e = e1 + e2`, if possible. -/
+def tryParseAdd (e : Q(Nat)) : Option (Q(Nat) × Q(Nat)) :=
+  match_expr e with
+  | HAdd.hAdd _ _ _ _ a b => some (a, b)
+  | Nat.add a b => some (a, b)
+  | _ => none
+
+/-- Cancel the leaf `n` out of `m`. Return `n'`, `m'` and a proof of `n' + m = n + m'`. -/
+partial def natCancelR (n m : Q(Nat)) : MetaM (Q(Nat) × Q(Nat) × Expr) := do
```
MackieLoeffel: Can we unify `natCancelR` and `natCancelL`?

alvinylt: Implementation size reduced by more than half upon code refactoring, with three theorems removed and six helper functions condensed into three.

Commit: [5dd6f4c](https://github.com/leanprover-community/iris-lean/pull/510/commits/5dd6f4ccd4106a38484b5641cb4a02434d043cac)

Iris/Iris/ProofMode/NatCancel.lean
alvinylt: Resolved as part of the code refactoring in commit [5dd6f4c](https://github.com/leanprover-community/iris-lean/pull/510/commits/5dd6f4ccd4106a38484b5641cb4a02434d043cac).

Iris/Iris/ProofMode/Tactics/ModIntro.lean
alvinylt: This function is actually obsolete as the current approach with `ipm_tactic_instance` is already able to handle the simplification of arithmetic expressions. It was only necessary previously because `NatCancel` was implemented in the same way as in stdpp using declarative type class instances. Changes to `imodintro` are thus reverted.

Commit: [2e0f73c](https://github.com/leanprover-community/iris-lean/pull/510/commits/2e0f73c2d65af4cbc8bb97e75faacf8e7a9a3335)

Iris/Iris/ProofMode/NatCancel.lean
alvinylt: Furthermore, some tests are added in `Tests/Instances.lean` to showcase how `NatCancel` handles various edge cases.

Commit: [011ec02](https://github.com/leanprover-community/iris-lean/pull/510/commits/011ec022d1eea40588393cbe1422cda5b1956ec4)

Iris/Iris/ProofMode/InstancesLater.lean
alvinylt: I have created a new branch (see [diff](https://github.com/ISTA-PLV/iris-lean/compare/4d8eee88a418f136700d3af2420de5c3091bcc87..iNextLaterCreditsAlt)) with an attempt to remove the `reducible` annotation and to revert all changes made to `IntoLaterN`.

However, this would result in `inext` failing to strip `▷?p`, as `laterIf` does not unfold to `laterN`, and `laterIf` itself is not registered as a modality. Here is an example:

```lean4
example [BI PROP] (p : Bool) (P Q : PROP) : ⊢ □ ▷ P -∗ □ ▷?p ▷ Q -∗ ▷?p ▷ □ (P ∗ Q) := by
  iintro #HP #HQ
  inext  -- imodintro: iprop(▷?p ▷ □ (P ∗ Q)) is not a modality matching iprop(▷^[?m.90]?m.91)
```

If we register a `Modality` instance for `▷?p` in terms of `IntoLaterN`, then it would defeat the purpose of simplifying the `IntoLaterN` instances. I'm unsure what alternative solutions are available though. Any suggestions?

MackieLoeffel: > If we register a Modality instance for ▷?p in terms of IntoLaterN, then it would defeat the purpose of simplifying the IntoLaterN instances.

I am not sure what you mean with this? I pushed https://github.com/ISTA-PLV/iris-lean/commit/cd6941234c9141c9ea10bea303db229f8a430026 to the branch which seems to make `inext` work by adding two instances, which does not seem too bad. However, there are still more instances missing since now we need to duplicate the instances for laterN also for laterIf. Not sure if this is good or bad. (I have to admit that I don't fully understand what is going on.)

alvinylt: > I pushed https://github.com/ISTA-PLV/iris-lean/commit/cd6941234c9141c9ea10bea303db229f8a430026 to the branch which seems to make inext work by adding two instances, which does not seem too bad. However, there are still more instances missing since now we need to duplicate the instances for laterN also for laterIf.

This looks reasonable to me, but I do agree that the need for duplicating the instances can be tedious.

The following is my understanding of the overall problem.

- The type class instance `into_laterN_laterN_bool` is not yet ported, so inext does not handle `▷?p` correctly. As an example, consider the following proof state:

  ```lean4
  ∗HP : ▷ P
  ∗HQ : ▷?p Q
  ⊢ ▷ (P ∗ Q)
  ```

  Upon applying `inext`, we get:

  ```lean4
  ∗HP : P
  ∗HQ : ▷^[p.toNat]Q
  ⊢ P ∗ Q
  ```

  The purpose of into_laterN_laterN_bool is to strip `▷?p` from a hypothesis when `inext` is used and the value of `p` is unknown. Therefore, in this example, we expect the proof state to be:

  ```lean4
  ∗HP : P
  ∗HQ : Q
  ⊢ P ∗ Q
  ```

  Currently, `BIBase.laterIf` is reducible, so `▷?p` unfolds to `▷^[p.toNat]` automatically. Therefore, the existing instance `intoLaterN_laterN` is applicable. Type class synthesis `intoLaterN_laterN` does not fail, and it simply leaves modality `▷?p` unchanged other than unfolding it as `▷^[p.toNat]`.
  
- As we try to port this instance in this PR, the realisation is that `into_laterN_laterN_bool` cannot be simply ported as a declarative instance. The instance `into_laterN_laterN_bool` has a lower priority than `intoLaterN_laterN`, and since synthesis with `intoLaterN_laterN` succeeds (without stripping `▷?p`), the instance `intoLaterN_laterN_bool` is never used.

- Assigning a higher priority for `intoLaterN_laterN_bool` does not solve the problem either. While in this case `intoLaterN_laterN_bool` is used, it tries to cancel `1` against `p.toNat` and still makes no progress.

- This is where the use of `MaybeIntoLaterN` along with `TCIf` and `TCEq` in the Rocq version becomes useful. The instance `intoLaterN_laterN_bool` has a lower priority than `intoLaterN_laterN`. It first tries `intoLaterN_laterN`, and if the type class search *succeeds* without making progress (by stripping any later modalities), `intoLaterN_laterN_bool` is then tried. The initial attempt in this PR was to port this approach directly by introducing `MaybeIntoLaterN` and `TCIf`, which solves the problem.

- The subsequent attempt in this PR (i.e. the latest revision on this branch) addresses the problem by replacing the instances `intoLaterN_later`, `intoLaterN_laterN` and `intoLaterN_laterN_bool` with a single `ipm_tactic_instance` that programmatically checks whether progress is made with `intoLaterN_laterN`. This approach also solves the problem, while also avoiding the need for `MaybeIntoLaterN` and other wrapper type classes.

alvinylt: The two solutions above mostly involve modifying `intoLaterN_later`, `intoLaterN_laterN` and `intoLaterN_laterN_bool`.

The alternative solution of setting `laterIf` irreducible also works, but in my opinion the consequence of having to duplicate all `laterN` instances may possibly make things harder to maintain, so the other two solutions are preferred.

MackieLoeffel:
The new design of `IntoLaterN` is very nice! I pushed some minor tweaks and I think this is ready to be merged. Thank you for all your work on this!

alvinylt:
Thanks for the review!

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -449,7 +449,7 @@ instance (priority := default - 10) intoLaterN_and_left [BI PROP]
 instance (priority := default - 11) intoLaterN_and_right [BI PROP]
     progress n (P P2 Q2 : PROP)
     [h : IntoLaterN (progress := true) (only_head := false) n P2 Q2] :
-    IntoLaterN progress (progress := false) n iprop(P ∧ P2) iprop(P ∧ Q2) where
```
alvinylt, line 452: Good catch. I was hoping that spelling `progress` and `only_head` out would rule out unintentional mismatches (and having the meaning of the Boolean values easier to read), but still missed this.

PR #511, opened 2026-07-12
feat: iframe with existential quantifiers

### LINKED ISSUE #390: Support existential quantifiers in iFrame

alvinylt:
## Description

Implements `iframe` with support for framing existentially quantified propositions.

Addresses #390 and #400.

Similar to `FrameInstantiateExistEnabled` in the Rocq version, this feature can be disabled for specific proofs/instances using the option `iris.frame.instantiateExists`. In the following example, `iframe` frames `HP` and `HS`. Without the `set_option iris.frame.instantiateExists false` line, it would solve the goal entirely.

```lean4
set_option iris.frame.instantiateExists false in
example [BI PROP] {α} (a : α) (P : PROP) (Q R : α → PROP) (S : PROP) :
    ⊢ P -∗ Q a -∗ R a -∗ S -∗ ∃ n, P ∗ Q n ∗ ∃ m, R m ∗ S := by
  iintro HP HQ HR HS
  iframe ∗
```

~~In the Rocq version, `FrameInstantiateExistEnabled` is used in `frame_wand`, `frame_forall`, `frame_impl_persistent` and `frame_impl`. The existing Lean counterparts of these four declarative type class instances are replaced by `SynthTactic` definitions so that the option `iris.frame.instantiateExists` can be set.~~

~~A caveat is that `SynthTactic` definitions have higher priority than declarative instances, while the instances `frame_wand`, `frame_forall`, `frame_impl_persistent` and `frame_impl` should have lower priority than `frame_here`. Now that they are defined as `SynthTactic` instances, we use the function `frameHereApplies` to specifically check whether `frame_here` is relevant and give priority to `frame_here` whenever possible.~~

~~We are using `mkAppM` instead of Qq in [`frameWp`](https://github.com/ISTA-PLV/iris-lean/blob/0c58c915614400a17e29d9fbd6cd9afbcf6113d1/Iris/Iris/ProgramLogic/WeakestPre.lean#L651) for similar reasons regarding `IProp GF` as in PR #510.~~

## Checklist


Iris/Iris/ProofMode/InstancesFrame.lean
```diff
@@ -353,3 +383,149 @@ def frameOr : SynthTactic := λ e => do
...
+    let .some instPers ← trySynthInstanceQ q(Persistent $P1) | return .continue
+    let .some instAbsorb ← trySynthInstanceQ q(QuickAbsorbing $P1) | return .continue
+    let Q2 : Q($prop) ← mkFreshExprMVarQ q($prop)
+    let some inst ← withFrameInstantiateExistsDisabled <|
+      synthInstanceRecursiveQ q(Frame false $R $P2 $Q2) | return .continue
+    return .success q(frame_impl $R $P1 $P2 $Q2 $instPers $instAbsorb $inst)
+
+@[ipm_tactic_instance Frame _ _ iprop(_ -∗ _) _]
+def frameWand : SynthTactic := fun e => do
+  let_expr Frame prop bi p R P _ := e | return .continue
```
MackieLoeffel: This is a lot more verbose than the previous version that is based on an instance. Would it be possible for the existential framing to check whether an assumption of a certain type is in the context, like in Rocq, such that we can use instances for wand and so on?

Iris/Iris/ProofMode/InstancesFrame.lean
```diff
@@ -353,3 +383,149 @@ def frameOr : SynthTactic := λ e => do
...
+    -- `iris.frame.instantiateExists` set as `true`
+    let a : Q($α) ← mkFreshExprMVarQ q($α)
+    let G : Q($prop) ← mkFreshExprMVarQ q($prop)
+    have body : Q($prop) := Expr.headBeta q($Φ $a)
+
+    let some inst ← synthInstanceRecursiveQ q(Frame $p $R $body $G)
+    | return .continue
+
+    let a ← instantiateMVars a
+    if !a.hasExprMVar then
```
MackieLoeffel: This check seems quite strong. Can you test what happens when there are mvars in `a`, but that not the mvars from the existential quantifier (both in Lean and in Rocq)?

Iris/Iris/ProofMode/InstancesFrame.lean
alvinylt: My original idea was that `set_option iris.frame.instantiateExists false in ...` and the monadic function `withFrameInstantiateExistsDisabled` could replace the job of `FrameInstantiateExistDisabled`.

I have now implemented `FrameInstantiateExistDisabled`. But instead of an empty type class as in Rocq, the relationship between `Frame` and `FrameInstantiateExistDisabled` in the Lean version is like the relationship between `AsEmpValid` and `AsEmpValid0`.

In the Rocq version, `FrameInstantiateExistDisabled` is used as a premise:

```rocq
Global Instance frame_wand p R P1 P2 Q2 :
  (FrameInstantiateExistDisabled → Frame p R P2 Q2) →
  Frame p R (P1 -∗ P2) (P1 -∗ Q2) | 2.
```

We could have defined `FrameInstantiateExistDisabled` as an empty type class in Lean and have something similar:

```lean4
instance frame_wand [BI PROP] p (R P1 P2 Q2 : PROP)
    [h : [FrameInstantiateExistDisabled] → Frame p R P2 Q2] :
    Frame p R iprop(P1 -∗ P2) iprop(P1 -∗ Q2) where ...
```

But for some reason when the synthesis for `Frame p R P2 Q2` is recursive, the premise `FrameInstantiateExistDisabled` is lost. As a result, `iframe` fails on various occasions.

So instead we have this:

```lean4
instance frame_wand [BI PROP] p (R P1 P2 Q2 : PROP)
    [h : FrameInstantiateExistDisabled p R P2 Q2] :
    Frame p R iprop(P1 -∗ P2) iprop(P1 -∗ Q2) where ...
```

With this approach, we are able to restore `frame_wand`, `frame_forall`, `frame_impl_persistent` and `frame_impl`. as declarative instances.

Iris/Iris/ProofMode/InstancesFrame.lean
alvinylt: Not sure whether the following is an example relevant to the question:

```lean4
example [BI PROP] (P Q : Nat → PROP) (m : Nat) :
    ⊢ P m -∗ (∃ n, Q n) -∗ ∃ x y : Nat, P x ∗ Q y := by
  iintro HP HQ
  iexists ?w
  iframe HP
  iassumption
```

Right before the use of `iframe`, the proof state is:

```
∗HP : P m
∗HQ : ∃ n, Q n
⊢ ∃ y, P ?w ∗ Q y
```

The metavariable `w` is instantiated with `m` by `iframe`, so after apply `iframe HP` we get:

```
∗HQ : ∃ n, Q n
⊢ ∃ y, Q y
```

Equivalent proof in Rocq:

```rocq
Lemma test_lemma (P Q : nat → PROP) (m : nat) : 
  P m -∗ (∃ n, Q n) -∗ ∃ x y : nat, P x ∗ Q y.
Proof.
  iIntros "HP HQ".
  iExists _.
  iFrame "HP".
  iAssumption.
Qed.
```

alvinylt: As `a` is the metavariable created for the existential quantifier binder, `a.hasExprMVar` checks whether it is instantiated after the recursive `Frame` synthesis for the existentially quantified body.

alvinylt: Found an example that might be worth looking into:

```lean4
example [BI PROP] (P : Option Nat → PROP) :
    ⊢ (∀ n, P (some n)) -∗ ∃ x, P x := by
  iintro HP
  ispecialize HP $$ %(?n)
  iframe HP
```

The current Lean implementation is unable to instantiate `x`, whereas a similar proof in Rocq does.

alvinylt: Fixed `frameExist` for properly handling of metavariables.

Commits: [142a286](https://github.com/leanprover-community/iris-lean/pull/511/commits/142a286cb86e24d475cc2617b1ea3e1d245ae58b) and [b9f5c05](https://github.com/leanprover-community/iris-lean/pull/511/commits/b9f5c057a1cf1a818cee09415950373db8c2956b)

MackieLoeffel:
@alvinylt I simplified the code in https://github.com/leanprover-community/iris-lean/pull/511/changes/0f9968b4ed71818ae27a031e61addbed0b0e0f29 Please have a look if this looks correct to you. If yes, I think this is ready to merge.

~~Edit: Something seems to be wrong since CI fails. I will look into this.~~ (fixed)

alvinylt:
The simplification of `solveGatherEvarsEq` looks good to me 👍🏼 

Regarding `frame_texist` and `frame_tforall`, I'm actually not so familiar with the various occurrences of telescopes in the code. In these two cases, telescopes refer to the defined in stdpp. But there is also `bi/telescopes.v` which is yet to be ported. Is it a Rocq-specific feature? Should `add_modal_tforall` in #500 also be ignored?

PR #401, opened 2026-05-20
feat: Port algebra/proofmode_classes.v (IsOp and related type classes)

alvinylt:
## Description

Ports [algebra/proofmode_classes.v](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/proofmode_classes.v) by implementing `IsOp`, its relevant type classes and type class instances.

Addresses #229.

Instances of these type classes exist in the following files but are yet to be ported.

- [x] [`algebra/lib/frac_auth.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/lib/frac_auth.v): `frac_auth_is_op`, `frac_auth_is_op_core_id`
- [x] [`base_logic/lib/own.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/lib/own.v): `into_sep_own`, `into_and_own`, `from_sep_own`, `combineSepAs_iOwn`, `from_and_own_persistent`
- [x] [`algebra/dfrac.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/dfrac.v): `dfrac_is_op`
- [x] [`algebra/view.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/view.v): `view_frag_is_op`, `view_auth_dfrac_is_op`
- [x] [`algebra/auth.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/auth.v): `auth_frag_is_op`, `auth_auth_dfrac_is_op`
- [x] [`algebra/frac.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/frac.v): `frac_is_op`, `is_op_frac`

The following files are yet to be ported but includes instances of `IsOp` or relevant type classes.

- [`algebra/gmap.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/gmap.v)
- [`algebra/numbers.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/numbers.v)
- [`algebra/lib/gmap_view.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/lib/gmap_view.v)
- [`algebra/lib/mono_list.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/lib/mono_list.v)
[...]

PR #513, opened 2026-07-13
feat: more expressive `semiOutParam` attribute for `ipm_class` declarations

alvinylt:
## Description

The [`InOut` datatype defined in `ProofModeM/SynthInstance.lean`](https://github.com/ISTA-PLV/iris-lean/blob/82ef884/Iris/Iris/ProofMode/SynthInstanceAttr.lean#L20-L22) is used along with `semiOutParam` in IPM type class declarations. Currently, we are reusing the attribute `semiOutParam` from the core Lean libraries (`Init.Prelude`), except that the semi-output parameters must be preceded by an `InOut` value so that it is clear whether particular type class instances require that parameter to be an input or an output.

The type class `IsOp`, introduced in PR #401, is an example where `semiOutParam` is used:

```lean4
class IsOp [CMRA α] (_ : InOut) (a : semiOutParam α) (_ : InOut) (b1 : semiOutParam α) (_ : InOut) (b2 : semiOutParam α)
```

We express `IsOp .in a .out b1 .out b2` for the splitting operation and `IsOp .out a .in b1 .in b2` for the merging operation. If we would like to use the type class for both directions, then currently the only way to do that is to parameterise all three `InOut` arguments. That is, `IsOp ioa a iob1 b1 iob2 b2` where `ioa`, `iob1` and `iob2` are arbitrary. However, this is quite inelegant as we know that `b1` and `b2` must always be the opposite of `a`.

My suggestion with this PR is to modify `semiOutParam` to remove the restriction that the `InOut` value must be the immediately preceding argument. Instead, we have the `semiOutParamIPM` attributes with two more explicit arguments to specify which `InOut` value to use. With `io : InOut := .in`, `semiOutParamIPM io α false` (or `semiOutParamIPM io α`) refers to an *input* parameter and `semiOutParamIPM io α true` refers to an *output* parameter. Likewise, with `io : InOut := .out`, `semiOutParamIPM io α false` (or `semiOutParamIPM io α`) refers to an *output* parameter and `semiOutParamIPM io α true` refers to an *input* parameter.

In fact, there is no longer any restriction for the `InOut` value to be a parameter in the type signature of the type class. In other words, we are able to use pattern matching, conditional branching and any reducible function to map other types of values to an `InOut` value.

With this design, `IsOp` is reformulated as follows, where `d : IsOp.Direction` is either `.merge` or `.split`.

```lean4
class IsOp [CMRA α]
    (d : IsOp.Direction) (a : semiOutParamIPM d.mapToInOut α)
    (b1 : semiOutParamIPM d.mapToInOut.negate α)
    (b2 : semiOutParamIPM d.mapToInOut.negate α) where
  is_op : a ≡ b1 • b2
```

There is then no need for one to think about the underlying details of type class synthesis when defining an `IsOp` instance.

There are six IPM type classes in the codebase that use semi-output parameters.
- `AsEmpValid`
- `FromWand`
- `IntoWand`
- `FromAssumption`
- `FromPure`
- `IsOp`

In fact, there is a similar situation with `AsEmpValid`, where the `PROP` and `bi` always practically have the same type class synthesis direction. As a result, it has also benefited from this PR:

```lean4
class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop)
    (PROP : semiOutParamIPM d.mapToInOut $ Type _)
    (bi : semiOutParamIPM d.mapToInOut $ BI PROP)
    (P : outParam $ PROP)
```

This is meant to just be a syntactic improvement, so there should not be any impact on the type class synthesiser's behaviour.  The core of this implementation is the type signature parsing in [`ProofModeM/SynthInstanceAttr.lean`](https://github.com/ISTA-PLV/iris-lean/blob/82ef884/Iris/Iris/ProofMode/SynthInstanceAttr.lean). Since we now have `semiOutParamIPM`, any use of `semiOutParam` in type class declarations with the `ipm_class` annotation is explicitly rejected to avoid confusion.

A caveat is that, for regular type class declarations (i.e., those without the `ipm_class` annotation), the attribute `semiOutParamIPM` is ignored. These two attributes are relevant to IPM type classes only and should not be used for regular, non-IPM type classes anyway.

As we are not reusing Lean's built-in `semiOutParam` attribute, a side effect is that the *built-in type class synthesiser* is unable to recognise that an argument is a semi-output parameter. This is a problem when an IPM type class is defined in terms of some non-IPM type classes. In this case, the type class synthesis for the non-IPM type classes by default uses the *built-in type class synthesiser*. Originally, with the `semiOutParam` attribute, Lean's built-in synthesiser is able to defer a premise whose type still mentions an unresolved semi-out parameter. This is why `synthInstanceMainCore` is slightly modified to handle the deferral explicitly.

## Checklist


Iris/Iris/ProofMode/Classes.lean
```diff
@@ -26,17 +26,29 @@ inductive AsEmpValid.Direction where
   | into
   | from
 
+meta section
+
+@[reducible]
+def AsEmpValid.Direction.mapToInOut (d : AsEmpValid.Direction) : InOut :=
```
MackieLoeffel: There is no mapping going on, so I would just call this `toInOut` (similar for the other function).

Iris/Iris/ProofMode/SynthInstance.lean
```diff
@@ -129,8 +129,20 @@ partial def synthInstanceMainCore (mvar : Expr) : MetaM (Option Unit) := do
...
+          let mut stillPending := []
+          for g in pending do
+            let mctxBefore ← getMCtx
+            if let some _ ← synthInstanceMainCore g then
+              progress := true
+            else
+              setMCtx mctxBefore
+              stillPending := stillPending.cons g
+          pending := stillPending
+        if !pending.isEmpty then return (none, true)
```
MackieLoeffel: It is quite annoying that this loop is necessary. Can we avoid this by making the new annotation sugar for something like `semiOutParam (inOutParam ...)` such that the semiOutParam is recognized by the lean typeclass machinery and the order of instances is computed correctly?

Iris/Iris/ProofMode/SynthInstanceAttr.lean
```diff
@@ -122,30 +128,44 @@ initialize ipmClassesExt :
...
+
+  One can use `InOut.negate` to negate the `InOut` value.
+
+  This should be used instead of `semiOutParam` for any type class with
+  the annotation `ipm_class`.
+-/
+@[reducible, expose]
+def semiOutParamIPM (_io : InOut) (α : Sort u) : Sort u := semiOutParam α
+
+/-- If `d` is `semiOutParam self α` or `semiOutParamNeg self α`, return `(negate, self)`. -/
```
MackieLoeffel: comment seems outdated.

Iris/Iris/ProofMode/SynthInstanceAttr.lean
```diff
@@ -62,7 +68,7 @@ checking synthesis order checking, which is useful (though not perfect).)
 inductive ParamKind
   | in
   | out
-  | semiOut
+  | semiOut (expr : Expr)
   | uncheckedIn
   | inout
```
MackieLoeffel: is inout still necessary?

Iris/Iris/ProofMode/SynthInstanceAttr.lean
```diff
@@ -164,27 +184,36 @@ def checkIPMSynthParams (type : Expr) : MetaM (Option (Array Nat)) := do
     let args := typeBody.getAppArgs
     if args.size != params.size then
       throwError "mismatched number of arguments"
-    let mut inoutIsIn := false
+
+    let mut resolved : Std.HashMap Nat Bool := {}
```
MackieLoeffel: What is this used for?

MackieLoeffel (commented):
I added some comments, but overall I think this is quite nice.

Iris/Iris/ProofMode/Classes.lean
alvinylt: Resolved in commit [b43d67e](https://github.com/leanprover-community/iris-lean/pull/513/commits/b43d67ebc4c785aec7be2af15c76903938236ed9).

Iris/Iris/ProofMode/SynthInstanceAttr.lean
alvinylt: Comment removed in commit [36527e4](https://github.com/leanprover-community/iris-lean/pull/513/commits/36527e4b86757a1d32920ddf5095652cbefbf69d).

Iris/Iris/ProofMode/SynthInstanceAttr.lean
alvinylt: No, it is now removed.

Commit: [4d25646](https://github.com/leanprover-community/iris-lean/pull/513/commits/4d256468500b11b9a35c640437f30543135d9b03).

Iris/Iris/ProofMode/SynthInstanceAttr.lean
alvinylt: Residual code from my previous approach. I originally had a more restrictive `semiOutParamIPM` where the `InOut` value is a dependent argument in the type class, but not necessary the preceding argument. This allowed multiple `semiOutParamIPM` parameters to refer to the same `InOut` argument.

This has subsequently been further generalised to allow arbitrary expressions of the `InOut` type. As a result, the mutable hash map is no longer relevant.

Removed in commit [4d25646](https://github.com/leanprover-community/iris-lean/pull/513/commits/4d256468500b11b9a35c640437f30543135d9b03).

Iris/Iris/ProofMode/SynthInstance.lean
alvinylt: Yes, I have now tried two alternative approaches. Both approaches avoid the need for the fixpoint loop.

The first approach (commit [30e665a](https://github.com/leanprover-community/iris-lean/pull/513/commits/30e665afa7f38f7520f5f7d308e2760b34323c60)) defines `semiOutParamIPM` as a macro for `semiOutParam (inOutParam ...)`. We throw an error when `semiOutParam` is used directly in an IPM type class declaration.

The second approach (commit [f905106](https://github.com/leanprover-community/iris-lean/pull/513/commits/f905106bdf58feb6c76d278dd2b110016b3c64ba)) is similar, except that we do not define the macro, and instead stipulate the direct use of `semiOutParam (inOutParam ...)` in IPM type class declarations.

---

Looks like [`computerSynthOrder` in `Lean/Meta/Instances.lean`](https://github.com/leanprover/lean4/blob/0d2a484/src/Lean/Meta/Instances.lean#L160-L162) is the cause: it uses `isAppOf` to syntactically look for `semiOutParam`. This is why a macro works but a reducible definition does not.

Also, the use of `semiOutParam` in both approaches triggers the built-in type synthesiser to check the synthesis order. This necessitates having `set_option synthInstance.checkSynthOrder false in ...` across the codebase. This would not be necessary with `semiOutParamIPM` being a `def`, so there is a tradeoff between the fixpoint loop and having `set_option synthInstance.checkSynthOrder false in ...`.

MackieLoeffel:
I made a small cleanup pass. Can you please resolve the merge conflicts? Afterwards, this is ready to merge.

alvinylt:
> Can you please resolve the merge conflicts?

Completed in commit [4ad6c81](https://github.com/leanprover-community/iris-lean/pull/513/commits/4ad6c811085ec7262e8d1bb248f5f2574ad199c8).

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -26,22 +26,37 @@ inductive AsEmpValid.Direction where
...
+def AsEmpValid.Direction.toInOut : AsEmpValid.Direction → InOut
+  | .into => .in
+  | .from => .out
+
+end
+
 @[ipm_class, rocq_alias AsEmpValid]
-class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) (_ : InOut) (PROP : semiOutParam $ Type _)
-(_ : InOut) (bi : semiOutParam $ BI PROP) (P : outParam $ PROP) where
+class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) io
```
MackieLoeffel, line 39: Why does `AsEmpValid` have an `io` parameter? Can it be computed from `d`?

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -26,22 +26,37 @@ inductive AsEmpValid.Direction where
...
-  (AsEmpValid.as_emp_valid io .out).2 rfl
+theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io}
+    (inst : AsEmpValid .from φ io PROP bi P) : (⊢ P) → φ :=
+  inst.as_emp_valid.right rfl
 
 @[ipm_class, rocq_alias AsEmpValid0]
-class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io1 : InOut) (PROP : semiOutParam $ Type _)
-    (io2 : InOut) (bi : semiOutParam $ BI PROP) (P : outParam PROP) where
-  as_emp_valid_0 : AsEmpValid d φ io1 PROP io2 bi P
+class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut := d.toInOut)
```
MackieLoeffel, line 56: Similar here, why is `io` necessary?

Iris/Iris/ProofMode/Classes.lean
alvinylt: Unfortunately we need that again for `istart` with BI instance specified. Previously, the synthesis direction for `PROP` and `bi` are both solely dependent on `d`. Now it is possible that `PROP` is an input parameter while `d = .from`.

I thought of having a function as follows, where the Boolean value input indicates whether the BI instance is specified in `istart`:

```lean4
def AsEmpValid.Direction.toInOutProp : AsEmpValid.Direction → Bool → InOut
  | .into, _ => .in
  | .from, true => .in
  | .from, false => .out
```

Do you prefer having this or having the `io` as it is?

MackieLoeffel: I see, this makes sense. Then let's keep the `io`.

PR #515, opened 2026-07-14
feat: port `algebra/stepindex.v`

alvinylt:
## Description

Port `algebra/stepindex.v`. It will be nice to have this ready before moving onto `algebra/stepindex_finite.v` as the latter requires refactoring a large number of files.

Addresses #203.

The current implementation defines `SIdx` as a type class that extends `LT`, `LE` and `Zero` from `Init.Prelude` so that we can reuse the notations `<`, `≤` and `0`. Meanwhile, `succᵢ` is a notation defined on our own.

<details>

<summary><b>Checklist of definitions and theorems</b></summary>

- [x] `SIdx.case`
- [x] `SIdx.eq_dec`
- [x] `SIdx.inhabited`
- [x] `SIdx.le_0_l`
- [x] `SIdx.le_0_r`
- [x] `SIdx.le_dec`
- [x] `SIdx.le_gt_cases`
- [x] `SIdx.le_lt_trans`
- [x] `SIdx.le_lteq` (ignored)
- [x] `SIdx.le_neq`
- [x] `SIdx.le_ngt`
- [x] `SIdx.le_po`
- [x] `SIdx.le_succ_diag_r`
- [x] `SIdx.le_succ_l`
- [x] `SIdx.le_succ_l_2` (ignored)
- [x] `SIdx.le_total`
- [x] `SIdx.limit`
- [x] `SIdx.limit_0`
- [x] `SIdx.limit_S`
- [x] `SIdx.limit_finite`
- [x] `SIdx.limit_lt_0`
- [x] `SIdx.lt_dec`
- [x] `SIdx.lt_ge_cases`
- [x] `SIdx.lt_le_incl`
- [x] `SIdx.lt_le_trans`
- [x] `SIdx.lt_nge`
- [x] `SIdx.lt_strict`
- [x] `SIdx.lt_succ_diag_r` (ignored)
- [x] `SIdx.lt_succ_diag_r'`
- [x] `SIdx.lt_succ_r`
- [x] `SIdx.lt_trans` (ignored)
- [x] `SIdx.lt_trichotomy` (ignored)
- [x] `SIdx.lt_wf` (ignored)
- [x] `SIdx.neq_0_lt_0`
- [x] `SIdx.neq_succ_0`
- [x] `SIdx.nlt_0_r` (ignored)
- [x] `SIdx.nlt_succ_r`
- [x] `SIdx.rec`
- [x] `SIdx.rec_lim`
- [x] `SIdx.rec_lim_ext` (ignored)
[...]

Issue #468, opened 2026-06-16
Deprecate Iris/Std/Classes.lean


markusdemedeiros:
The classes seem to be sparsely used now. We should look and see if they can be removed.

alvinylt:
It seems like most of these type classes are already defined in [`Init.Core`](https://github.com/leanprover/lean4/blob/8c9756b/src/Init/Core.lean#L2562-L2595).

I have started porting step-indexing for Transfinite Iris. There are some type classes (e.g. irreflexivity, strict order, partial order, antisymmetry) that are not available in `Iris/Std/Classes.lean`. So we'll have to decide whether to introduce the missing ones into the file or reuse built-in definitions.

Are there particular reasons to prefer `Iris/Std/Classes.lean` over `Init.Core`? If not, shall we use the latter throughout the codebase for consistency?

alvinylt:
Also, the built-in type class `Trans` (for transitivity) enables `calc`-style proofs. Currently [`BIBase.Entails.trans`](https://github.com/leanprover-community/iris-lean/blob/5a790aed0dcad30219aa218c88a7fd306c092fae/Iris/Iris/BI/BI.lean#L87-L88) is a theorem expressed in terms of `Iris.Std.Transitive` while [`entails_trans` in `DerivedLaws.lean`](https://github.com/leanprover-community/iris-lean/blob/5a790ae/Iris/Iris/BI/DerivedLaws.lean#L24-L25) is an instance of `Trans`.

As a side note we can perhaps also annotate the reflexivity theorem with `@[refl]`:

```lean4
@[simp, refl]
theorem BIBase.Entails.rfl [BI PROP] {P : PROP} : P ⊢ P := refl
```

This allows us to use the tactic `rfl`. Example:

```lean4
example [BI PROP] (P : PROP) : P ⊢ P := by rfl
```

as an alternative to `.rfl` or `by exact .rfl`.

lzy0505:
We should use the built-in typeclasses throughout.

PR #518, opened 2026-07-15
refactor: remove duplicate type classes in `Iris/Std/Classes.lean` and reuse definitions from core libraries

alvinylt:
## Description

Addresses #468.

This PR involves four commits, one for each set of changes.
- [The first commit](https://github.com/leanprover-community/iris-lean/commit/e14416646ada0ce9033c8aae7efe8014b414bf05) removes the type classes `Reflexive` and `Transitive` from `Iris/Std/Classes.lean` and replaces them with [`Refl` in `Init/Core.lean`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2563-L2566) and [`Trans` in `Init/Prelude.lean`](https://github.com/leanprover/lean4/blob/4333825/src/Init/Prelude.lean#L1344-L1348), respectively.
  - The custom type class `Preorder` remains in `Iris/Std/Classes.lean` and adjusted to use the built-in type classes.
  - By annotating the theorem [`BIBase.Entails.rfl` in `BI.lean`](https://github.com/leanprover-community/iris-lean/blob/e144166/Iris/Iris/BI/BI.lean#L90) with `@[refl]`, we are able to use the tactic `rfl` to tackle goals in the form `P ⊢ P` as an alternative to `by apply .rfl`.

- [The second commit](https://github.com/leanprover-community/iris-lean/commit/69302f391dddaa465ba643e1067e0651535bc8e1) eliminates the custom type class `Preorder` and replaces it with [`IsPreorder` in `Init/Data/Order/Classes.lean`](https://github.com/leanprover/lean4/blob/efbd4b8/src/Init/Data/Order/Classes.lean#L50-L56). In fact, the type class `IsPreorder` is relatively new (see [Lean 4.25.0 documentation](https://lean-lang.org/doc/reference/latest/releases/v4.25.0)).
  - In order for the type class `IsPreorder` to work for `BI PROP`, there must be an instance of `LE PROP`. At this point my approach was to extend the type class `BIBase`:
    ```lean4
    class BIBase (PROP : Type u) extends LE PROP
    ```
    But this is not really ideal, as the BI entailment (`⊢`) is now *definitionally equal* to `≤` in the type class `LE`, which means `simp` may possibly unfold `⊢` as `≤`, and the `@[refl]` annotation introduced in the first commit would not achieve its purpose because `rfl` would require the `@[refl]` annotation directly for `≤`.

- [The third commit](https://github.com/leanprover-community/iris-lean/commit/af17f95e6a6ad2e5b1812a564e03520b50e6066d) addresses the problem with `LE PROP` in a different way.
  - Instead of modifying `BIBase` and having `Std.IsPreorder PROP` as a field of `BI`, we instead have the bare propositions of reflexivity and transitivity as two fields of `BI`:
    ```lean4
    entails_refl {P : PROP} : P ⊢ P
    entails_trans {P Q R : PROP} : (P ⊢ Q) → (Q ⊢ R) → P ⊢ R
    ```
  - We then prove that there exist `LE PROP` and `IsPreorder PROP` as separate instances in terms of `entails_refl` and `entails_trans`. This is more consistent with how other Lean code is organised.
  - This reverts the definition `⊢` as `≤` in the previous commit.

- [The fourth commit](https://github.com/leanprover-community/iris-lean/commit/96462afd66946e7dd408632bb3b66c81517ffada) eliminates `Iris/Std/Classes.lean` entirely.
  - There is a number of type class instances in `BI/DerivedLaws.lean` using definitions from `Iris/Std/Classes.lean`. There are similar type classes for commutativity, associativity, antisymmetry, `LeftAbsorb`, `RightAbsorb`, `LeftId` and `RightId` in [`Init/Core.lean`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2473-L2598), but they are expressed in terms of equality rather than bi-entailment.
  - I would suggest leaving these type class instances in `BI/DerivedLaws.lean` out until the built-in type classes are compatible with this use case, rather than using a mixture of built-in type classes and custom, more generic type classes. (Perhaps ongoing development regarding bi-entailment as equality might help?)
  - `Iris/Std/Classes.lean` defines the notation `⊤` for the top element for instances of the type class `Top`. This is used only once in [`CoPset.lean`](https://github.com/leanprover-community/iris-lean/blob/96462af/Iris/Iris/Std/CoPset.lean#L255-L256), so I moved to the notation there.
  - ~~Similarly, the notation `##` for disjointness is used only once in [`GenSets.lean`](https://github.com/leanprover-community/iris-lean/blob/96462af/Iris/Iris/Std/GenSets.lean#L72-L73) and thus defined there directly.~~ The notation `##` for disjointness is used in [`GenSets.lean`](https://github.com/leanprover-community/iris-lean/blob/96462af/Iris/Iris/Std/GenSets.lean#L72-L73) and [`Namespaces.lean`](https://github.com/leanprover-community/iris-lean/blob/96462af/Iris/Iris/Std/Namespaces.lean#L40-L41).
  
### Remarks
    
Step-indexing (PR #515) further involves irreflexivity, total relations and strict orders. At the moment, I have introduced these three new definitions into `Iris/Std/Classes.lean`. If we decide to deprecate `Iris/Std/Classes.lean`, we can directly use [`irrefl`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2588-L2591) and [`total`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2583-L2586) from `Init/Core.lean`, as well as [`IsPartialOrder`](https://github.com/leanprover/lean4/blob/master/src/Init/Data/Order/Classes.lean#L63-L64) from `Init/Data/Order/Classes.lean`. Not aware that a type class for strict orders (irreflexivity + transitivity) exists though.

Would be great to discuss whether to adopt any of these changes.

## Checklist


Iris/Iris/Std/Rewrite.lean
```diff
@@ -135,7 +134,7 @@ where
 
   applyReflexivity (goal : MVarId) : TacticM Unit := do
     try
-      discard <| apply' goal ``refl
+      discard <| apply' goal ``Std.Refl.refl
```
alvinylt: Relevant to the discussions: https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean

Iris/Iris/Std/Classes.lean
```diff
@@ -1,81 +0,0 @@
...
-
-/-- Require that an element `i` of `α` is the left unit of a binary function `f` on `α` in a
-relation `R` on `α`. -/
-class LeftId (R : Relation α) (i : α) (f : α → α → α) where
-  left_id {x : α} : R (f i x) x
-export LeftId (left_id)
-
-/-- Require that an element `i` of `α` is the right unit of a binary function `f` on `α` in a
-relation `R` on `α`. -/
-class RightId (R : Relation α) (i : α) (f : α → α → α) where
```
MackieLoeffel, line 54: Please keep the classes where there is no replacement provided by Lean.

alvinylt: Keeping only these changes:
- Removing `Reflexive` and `Transitive`
- Remove `Preorder`, use `Std.IsPreorder` instead
- Improving the definition of `BI`

Commit: [65273e6](https://github.com/leanprover-community/iris-lean/pull/518/commits/65273e69de26165128ec1a82d86c1e8c77d9e3df)

Iris/Iris/Std/Classes.lean
```diff
@@ -19,19 +19,7 @@ export Top (top)
...
-
-/-- Require that a relation `R` on `α` is transitive. -/
-class Transitive (R : Relation α) where
-  trans {x y z : α} : R x y → R y z → R x z
-export Transitive (trans)
-
-/-- Require that a relation `R` on `α` is a preorder, i.e. that it is reflexive and transitive. -/
-class Preorder (R : Relation α) extends Reflexive R, Transitive R
-
+def trans [Trans R R R] : R a b → R b c → R a c := Trans.trans
```
markusdemedeiros: Should we just make clients use `Trans.trans` instead?

Iris/Iris/BI/MonPred.lean
```diff
@@ -43,8 +43,11 @@ predicates are functions out of a `BiIndex` that are monotone w.r.t. this order.
 structure BiIndex where
   car : Type _
   [inhabited : Inhabited car]
-  rel : car → car → Prop
-  [preorder : Std.Preorder rel]
+  -- rel : car → car → Prop
```
markusdemedeiros: formatting: remove spaces and comments

markusdemedeiros (changes requested):
Looks good by me, with two small comments.

Iris/Iris/BI/MonPred.lean
alvinylt: Resolved in commit [a95d97d](https://github.com/leanprover-community/iris-lean/pull/518/commits/a95d97dd76fcf5568c52500c58be6b15a0365879).

Iris/Iris/Std/Classes.lean
alvinylt: Yes, the theorem is now removed.

Commit: [57563f6](https://github.com/leanprover-community/iris-lean/pull/518/commits/57563f631539d792b732294b3f422474d8a8484a)

alvinylt:
For future reference, here is the summary of the changes.

| Type class in `Iris/Std/Classes.lean`                                                                                        | Changes                                                                                                                                                   | Remarks                                                                                                                                                                                                                                                |
|------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `Reflexivity`                                                                                                                | Replaced by [`Refl` from `Init/Core.lean`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2563-L2566)                               | Exact same definition.                                                                                                                                                                                                                                 |
| `Transitivity`                                                                                                               | Replaced by [`Trans` from `Init/Prelude.lean`](https://github.com/leanprover/lean4/blob/4333825/src/Init/Prelude.lean#L1344-L1348)                        | Instances of `Trans` enable `calc`-style proofs. The built-in type class `Trans` is also more generic than `Transitivity` as the former is parametrised over three relations.                                                                          |
| `Preorder`                                                                                                                   | Replaced by [`IsPreorder` from `Init/Data/Order/Classes.lean`](https://github.com/leanprover/lean4/blob/efbd4b8/src/Init/Data/Order/Classes.lean#L50-L56) | Useful for `grind`.                                                                                                                                                                                                                                    |
| `Commutative`, `Associative`, `Idempotent`, `LeftId`, `RightId`, `LeftAbsorb`, `RightAbsorb`, `Associative`, `Antisymmetric` | None                                                                                                                                                      | Similar to the type classes in [`Init/Core.lean`](https://github.com/leanprover/lean4/blob/cc9a8de/src/Init/Core.lean#L2473-L2598) but those are not generic enough as they are formulated in terms of equality (`=`) rather than any binary relation. |
| `Top`                                                                                                                        | None                                                                                                                                                      | Same definition available in Mathlib ([`Mathlib/Order/Notation.lean`](https://github.com/leanprover-community/mathlib4/blob/44ea3a0e13339f31231dde209c4f8196e179065d/Mathlib/Order/Notation.lean#L180-L184)) but not in the core libraries.            |
| `Disjoint`                                                                                                                   | None                                                                                                                                                      | Available as a definition in Mathlib ([`Mathlib/Order/Disjoint.lean`](https://github.com/leanprover-community/mathlib4/blob/44ea3a0e13339f31231dde209c4f8196e179065d/Mathlib/Order/Disjoint.lean#L45-L49)) rather than a type class.                                   |

markusdemedeiros (approved):

PR #523, opened 2026-07-18
feat: texan triple minor improvements

ayhon:
## Description
- Corrects a minor typo in `BI/WeakestPre.lean`
- Disallows writing `{ { P }} e {{ Q }}`, that is, whitespace between the `{` for both texan triples and weakest preconditions
- Fixes the definition for texan triples, which was using `→` instead of `-∗`
- Allows, in the definition of texan triples, to specify the types of the newly introduced binders (basically, it follows more closely the type of syntax nodes that are allowed as binders in a `forall`)


markusdemedeiros:
Nice fixes, thanks!

PR #525, opened 2026-07-19
feat: add support for `match` in `iprop(...)`

alexkeizer:
## Description

This PR adds a macro expansion, such that in `iprop(match ... with ...)` the right-hand-side of each match alternative is wrapped with `iprop` again.

For example, this enables the parsing of the following:
```lean4
iprop(match p with
  | true => P ∗ Q
  | false => Q ∗ P)
```

## Checklist

## Generative AI Guidelines
This code is 100% human-written

### DIFF
--- a/Iris/Iris/BI/Notation.lean
+++ b/Iris/Iris/BI/Notation.lean
@@ -34,7 +34,7 @@ macro_rules
               $[$alts:matchAlt]*)) => do
         let alts ← alts.mapM <| fun
           | `(matchAltExpr| | $[$lhs]|* => $rhs) => `(matchAltExpr| | $[$lhs]|* => iprop($rhs))
-          | alt => return ⟨alt⟩
+          | _ => throwUnsupported
         `(match $[$g:generalizingParam]? $[$m:motive]? $[$x:matchDiscr],* with $[$alts:matchAlt]*)
 
 macro:max "iprop(" P:term " : " t:term ")" : term => `((iprop($P) : $t))
--- a/Iris/Iris/BI/Notation.lean
+++ b/Iris/Iris/BI/Notation.lean
@@ -1,16 +1,18 @@
 /-
 Copyright (c) 2022 Lars König. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Lars König
+Authors: Lars König, Alex Keizer
 -/
 module
 
 public meta import Lean.PrettyPrinter.Delaborator
 
+meta import Lean.Parser.Term
+
 public meta section
 
 namespace Iris.BI
-open Lean Lean.Macro
+open Lean Lean.Macro Lean.Parser.Term
 
 -- define `iprop` embedding in `term`
 syntax:max "iprop(" term ")" : term
@@ -27,6 +29,13 @@ macro_rules
   | `(iprop(if $c then $t else $e)) => ``(if $c then iprop($t) else iprop($e))
   | `(iprop(($P : $t)))             => ``((iprop($P) : $t))
   | `(iprop(fun $xs* => $P))        => ``(fun $xs* => iprop($P))
+  -- `iprop(match …)` expansion wraps the rhs of each match arm in `iprop(…)`
+  | `(iprop(match $[$g:generalizingParam]? $[$m:motive]? $[$x:matchDiscr],* with
+              $[$alts:matchAlt]*)) => do
+        let alts ← alts.mapM <| fun
+          | `(matchAltExpr| | $[$lhs]|* => $rhs) => `(matchAltExpr| | $[$lhs]|* => iprop($rhs))
+          | alt => return ⟨alt⟩
+        `(match $[$g:generalizingParam]? $[$m:motive]? $[$x:matchDiscr],* with $[$alts:matchAlt]*)
 
 macro:max "iprop(" P:term " : " t:term ")" : term => `((iprop($P) : $t))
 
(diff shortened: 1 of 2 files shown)


Iris/Iris/BI/Notation.lean
```diff
@@ -27,6 +29,13 @@ macro_rules
   | `(iprop(if $c then $t else $e)) => ``(if $c then iprop($t) else iprop($e))
   | `(iprop(($P : $t)))             => ``((iprop($P) : $t))
   | `(iprop(fun $xs* => $P))        => ``(fun $xs* => iprop($P))
+  -- `iprop(match …)` expansion wraps the rhs of each match arm in `iprop(…)`
+  | `(iprop(match $[$g:generalizingParam]? $[$m:motive]? $[$x:matchDiscr],* with
+              $[$alts:matchAlt]*)) => do
+        let alts ← alts.mapM <| fun
+          | `(matchAltExpr| | $[$lhs]|* => $rhs) => `(matchAltExpr| | $[$lhs]|* => iprop($rhs))
+          | alt => return ⟨alt⟩
```
alexkeizer: The first alternative *should* capture all forms of match alternatives parse-able in Lean today, but in case there's any syntax that doesn't get recognized I choose to just pass the rhs through unchanged in the second arm, as opposed to throwing a "unrecognized syntax" error.

Iris/Iris/BI/Notation.lean
```diff
@@ -48,5 +57,16 @@ partial def unpackIprop [Monad m] [MonadRef m] [MonadQuotation m] : Term → m T
...
+      -- The following type ascriptions look redundant, but, without them, the ``(match ...)`
+      -- syntax quotation below fails with an error about types containing metavariables.
+      let g : Option (TSyntax ``generalizingParam) := g
+      let mot : Option (TSyntax ``motive) := mot
+      let alts ← Array.mapM (as := alts) (m:=m) <| fun
+        | `(matchAltExpr| | $[$lhs]|* => $rhs) => do
+            let rhs ← unpackIprop rhs
+            `(matchAltExpr| | $[$lhs]|* => $rhs)
+        | alt => return ⟨alt⟩
+      `(match $[$g:generalizingParam]? $[$mot:motive]? $[$x:matchDiscr],* with $[$alts:matchAlt]*)
```
alexkeizer, line 69: I'm not entirely sure what `unpackIprop` is meant to do, or how to test it, but looking at the other cases (e.g, `if then else`) it seems to be recursively called on all sub-expression that we wrapped in `iprop` in the macro expansion, so that is what I implemented for `match` also.

markusdemedeiros:
Nice. I'm glad that there's a way to do this without writing our own matching primitive.

Iris/Iris/BI/Notation.lean
MackieLoeffel: Can we throw an error here that says that having an unparseable match is a bug that should not happen?

Iris/Iris/BI/Notation.lean
MackieLoeffel: In principle this should be the correct thing to do, but it does not seem to fully work at the moment:
```
/--
info: iprop(∀ p,
    True ∗
      match p with
      | true => Ψ 1
      | false => iprop(False)) : PROP
-/
#guard_msgs in #check iprop(∀ p, True ∗ match p with
  | true => Ψ 1
  | false => False)
```
In the output of this test `iprop(False)` should be `False`.

MackieLoeffel (commented):
Great, thanks for the PR! I responded to the comments.

Iris/Iris/BI/Notation.lean
alexkeizer: Yes, this is pretty straightforward, I've now changed this to throw an [`unsupportedSyntax`](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Lean.Macro.Exception.unsupportedSyntax) exception!

Iris/Iris/BI/Notation.lean
alexkeizer: Looking at other tests in that file, this behaviour seems to match how if-then-else gets printed:
````lean
/-- info: if true = true then Ψ 1 else iprop(False) : PROP -/
#guard_msgs in #check iprop(if true then term(Ψ 1) else False)
````

I believe getting the output you want might be a bit difficult.

To start, the wrong output you highlight above is not because `unpackIprop` does the wrong thing, rather, `unpackIprop` never gets called the produce the pretty-printed / delaborated output. This is because delaborators are keyed on whatever function constant is being applied in the expression to be pretty-printed.
For example, for an iris connective like `□`, `iprop(□ P)` macro-expands to `intuitionistically iprop(P)`, and this constant has a corresponding `delab_rule` defined, which causes Lean to pretty print `intuitionistically P` in terms of `iprop(□ ...)`:
```
delab_rule intuitionistically
  | `($_ $P) => do ``(iprop(□ $(← unpackIprop P)))
```

Unfortunately, in the `match` test case you bring up, this does not expand to an Iris constant being applied, `iprop(match ...)` just produces a regular Lean match construct. I *think* we cannot override the delaboration of matches, but even if we could, it would probably be a bad idea (since we would be overriding pretty printing of *all* matches, not just `iprop` produced ones). Idem for the similarly behaving if-then-else test case.



That all being said, we *can* test `unpackIprop` by having the match not be at the top-level as in:
```lean
/--
info: iprop(□
    match p with
    | true => Ψ 1
    | false => False) : PROP
-/
#guard_msgs in #check iprop(□ match p with
  | true => term(Ψ 1)
  | false => False)
```
which was actually broken before, but I've now pushed a commit to fix. 

(At least, this is my understanding from digging through your code-base and a cursory knowledge of Lean's pretty-printing API; feel free to point out if I've misunderstood) 

How much of a blocker is this delaboration output? Given that `iprop(match ...)` matches the behaviour of `iprop(if ... then ... else ...)`, and the complexity likely needed to solve this "properly", would you be OK with merging this as is?

alexkeizer:
@MackieLoeffel I responded to your comments, could you have another look?

Iris/Iris/BI/Notation.lean
MackieLoeffel: Yes, this is what I had in mind. It seems indeed not possible to change the delaboration for top-level `match` and `if`, but it is good that you fixed the nested case.

MackieLoeffel (approved):
This looks great now, thank you!

Iris/Iris/BI/Notation.lean
alexkeizer: Thinking about this a bit more. Technically, it *would* possible to define some kind of tagging identity function, like so:
```lean
abbrev ppAsIprop {a} (x : a) : a := x
```
Then, have `iprop(match ...)` produce `ppAsIprop (match ...)` and register a custom delaborator for `ppAsIprop` that produces the output we want. With this being a reducible definition, things like `simp` and `grind` should unfold it as needed, so it *should* be transparent.

That said, it definitely feels a bit magical to me, and having such a function in the expression but hidden from the info-view definitely has potential for unintended confusion, so I don't think we should implement this at the moment, but only consider it again if we have a strong justification for needing the "prettier" printing.

alexkeizer:
Thanks for the quick reviews!

PR #532, opened 2026-07-23
refactor: use simp_to_model for TreeMap mergeWith

### LINKED ISSUE #127: Refactor TreeMap to use `simp_to_model`
The standard library has a tactic `simp_to_model` for proving facts about TreeMaps. We prove some facts about TreeMaps in `HeapInstances.lean`, but we do not make use of this automation yet. To make these proofs more robust, we should figure out how.

ctkrug:
## Description

Replace the bespoke `TreeMap.mergeWith` lookup proof and its private fold helpers with a direct `simp_to_model` proof. This preserves both public theorem signatures while replacing 122 lines of implementation-specific machinery with a 10-line model-based proof.

This draft depends on the Std model bridge in leanprover/lean4#14526.

Validation performed on the exact committed diff:

- native Lean v4.32 release build with the proposed Std bridge
- focused downstream `simp_to_model` bridge checker
- full Iris-Lean build (259 jobs)
- independent theorem-signature and proof-safety checker

Fixes #127.

## Checklist

* [x] My code follows the mathlib naming and code style conventions
* [x] No file-level authors list needs updating for this proof refactor

## Generative AI Guidelines

AI assistance was used to investigate and draft the refactor. The final proof was manually reviewed, the duplicated constructions were removed, both public theorem signatures were preserved, and all checks listed above passed.

### DIFF
--- a/Iris/Iris/Std/HeapInstances.lean
+++ b/Iris/Iris/Std/HeapInstances.lean
@@ -84,32 +84,6 @@ end ClassicalAllocHeap
 
 end Iris.Std
 
-section Lemmas
-
-/-- Merge an optional value with an optional key-value pair, using the pair's key in the
-merge function. This is an internal helper for TreeMap heap proofs. -/
-private def Option.pairMerge (f : K → V → V → V) (o1 : Option V)
-    (o2 : Option (K × V)) : Option V :=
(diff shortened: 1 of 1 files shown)


markusdemedeiros:
Awesome! Thanks for figuring this out.

PR #537, opened 2026-07-25
feat: add unbounded heap for ExtTreeMap

### LINKED ISSUE #126: Improve TreeMap Heaps
TreeMap heaps currently do not support allocation. Instanstate `UnboundedHeap` for them in `HeapInstances.lean`.

The challenge in this is going from a TreeMap (which has finite domain) to a key that does not exist in the map. If you can do this, it becomes your `fresh` function. Like the AssocList heap, your `NotFull` predicate should be `True`.

kernelpanic888:
## Description

Closes #126.

Instantiate `UnboundedHeap` for `ExtTreeMap K · compare` when the key type is infinite. The fresh key is chosen outside the finite key projection of the map's lawful `toList`; `LawfulFiniteMap.toList_get` then proves that lookup at this key is `none`.

As requested in the issue, `notFull` is `True`, so emptiness and preservation after fresh insertion are immediate.

## Validation

- `lake build Iris.Std.HeapInstances`
- `lake build` (260 jobs)
- `git diff --check`

## Checklist

- [x] The change follows the project's Lean naming and style conventions.
- [x] No authors-list update is appropriate for this focused instance addition.

### DIFF
--- a/Iris/Iris/Std/HeapInstances.lean
+++ b/Iris/Iris/Std/HeapInstances.lean
@@ -281,32 +281,24 @@ instance : LawfulFiniteMap (ExtTreeMap K · compare) K where
     exact List.pairwise_map.mpr distinct_keys_toList
   toList_get {_ m _ _} := m.mem_toList_iff_getElem?_eq_some
 
-noncomputable instance instUnboundedHeapCompare [InfiniteType K] :
+end HeapInstance
+end Std.ExtTreeMap
+
+namespace Classical
+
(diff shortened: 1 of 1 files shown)


markusdemedeiros:
Good catch! This instance should exist. I made some changes:
- `ExtTreeMap` already has a way to get the list of keys
- Much of the proof could have been shortened by grind
- Because this instance includes data that does not evaluate, it should be scoped to classical.

Issue #206, opened 2026-04-07
Port bi/lib/counterexamples.v


lzy0505:
## Summary

Port [`bi/lib/counterexamples.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/bi/lib/counterexamples.v).

## Dependencies

**Rocq dependencies:**

- `bi/bi` (aggregation)
- `proofmode/proofmode` (aggregation)

PR #540, opened 2026-07-26
feat: port `bi/lib/counterexamples.v` (+ fix `IntoWand` instances)

alvinylt:
## Description

Ports `bi/lib/counterexamples.v`.

Addresses #206.

Each `Module` in the Rocq version corresponds to a `namespace` in `BI/Lib/Counterexamples.lean`.
- [x] `affine_em`: `AffineEM`
- [x] `löb_em`: `LoebEM`
- [x] `savedprop`
- [x] `inv`
  - [x] `inv1`
  - [x] `inv2`
- [x] `linear`
- [x] `later_credits_plain`

### Fixing `IntoWand` Instances

There is a comment above the existing type class instances `intoWand_imp_false` and `intoWand_imp_true`:

> *TODO: compare this with `into_wand_impl_false_false`, `into_wand_impl_false_true`, ... in Rocq*

At one point within the proof of `later_anything`, we come across the following proof state:

```lean4
□HnotP : ¬▷ False
□IH : ▷ False
⊢ False
```

With the existing two instances, we are unable to use `ispecialize HnotP $$ IH` to solve the goal.

The instance `intoWand_imp_true` stipulates `Affine (▷ False)`, which does not hold. The type class instance [`into_wand_impl_true_true`](https://gitlab.mpi-sws.org/iris/iris/-/blob/a3a21bd4/iris/proofmode/class_instances.v#L450-L462), on the other hand, has no such requirement. The instances `intoWand_impl_false_true` and `intoWand_impl_false_false` also come with different assumptions.

The two instances are therefore replaced by those directly corresponding to the formulation in Rocq.

## Checklist


Iris/Iris/BI/Lib/Counterexamples.lean
```diff
@@ -0,0 +1,264 @@
...
+  icases affinely_elim $$ Hfalse with Hfalse
+  iinduction n with
+  | zero =>
+    dsimp [BIBase.laterN, Nat.repeat]
+    iexfalso; iexact Hfalse
+  | succ n IH =>
+    ihave Hfalse := Hfalse
+    /- Necessary to unfold `▷^[n + 1]` as `▷ ▷^[n]`, or else we get
+       `∗Hfalse : ▷^[n + 1] False` after `imod`. -/
+    icases (later_laterN n).mp $$ Hfalse with Hfalse
```
alvinylt: We have to unfold `laterN` manually in order for `imod` to produce the correct hypothesis.
Might be interesting to investigate the cause.

Iris/Iris/BI/Lib/Counterexamples.lean
```diff
@@ -0,0 +1,573 @@
+/-
+Copyright (c) 2026 Alvin Tang. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Alvin Tang
+-/
+module
+
+public import Iris.BI
+public import Iris.ProofMode
```
MackieLoeffel, line 9: Can you please add some comments describing the counter examples?

Iris/Iris/BI/Lib/Counterexamples.lean
MackieLoeffel: One can fix this by adding a `FromAssumption` instance. 
```
instance fromAssumption_later_laterN [BI PROP] (p : Bool) (P Q : PROP)
    [h : FromAssumption p .in iprop(▷^[n]P) Q] : FromAssumption p .in iprop(▷^[n+1]P) iprop(▷ Q) where
  from_assumption := by sorry
```
Alternatively, just using `simp [(later_laterN n).to_eq]` also works and might be less magic. 
Not sure what is the better option.

Iris/Iris/BI/Lib/Counterexamples.lean
alvinylt: Comments added in commit [de49191](https://github.com/leanprover-community/iris-lean/pull/540/commits/de49191b4bfb5ecb8c2a5363bb6a1420042efda8), mostly based on what is already in the Rocq version.

alvinylt:
> One can fix this by adding a FromAssumption instance.

I have added the instance, and it fixes the problem. I reckon this is a nice solution as it avoids the need for the manual unfolding of `laterN`. Do you know if this is specifically due to how inductive data structures are unfolded in Lean? This issue occurs even if we define `laterN` inductively without using `Nat.repeat`. There seems to be no similar issue in Rocq.

Commit: [932b781](https://github.com/leanprover-community/iris-lean/pull/540/commits/932b78174141fe0bfa43bb5ebabedba026abdf8d)

MackieLoeffel:
Rocq does not need the instance since there `▷^[n + 1]` is convertible to `▷ ▷^[n]` in Rocq, but not in Lean. I've added a comment about this. Thanks for the PR! Merging.

PR #549, opened 2026-07-28
cleanup: post-setoid proof cleanup

Kaptch:
## Description
A pass over proofs that relied on distance, where they could use equality. 

I've added some private lemmas BI/Algebra.lean to simplify some ugly proofs there, but probably these lemmas can be made public and moved to Sbi (probably they can be used to define logic over other algebras as well, since they are just for Sbi juggling).


MackieLoeffel:
@Kaptch Are you planning to finish this PR or should it be closed?

Kaptch:
> @Kaptch Are you planning to finish this PR or should it be closed?

Yes, I'll bump and finish it. It won't grow much larger from the current state.

PR #552, opened 2026-07-29
chore: imports in `Iris.lean`

alvinylt:
## Description

I assume [`Iris.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris.lean) is supposed to import all modules. While I was working on the implementation of IPM tactics, there were multiple occasions where the changes broke something in the codebase, but the error did not show up when building `Iris.lean` and were only later flagged by CI. It turned out that those files were not included by `Iris.lean`, so this PR adds those missing entries.

## Checklist


alvinylt:
As a side note, certain modules such as `Iris.Std.RocqPorting` are currently imported directly by many files. Meanwhile there are also plenty that do so transitively. On another occasion during the implementation of `iinv`, IPM tactics have been useful for proving type class instances in `ProofMode/Instances.lean` (e.g. [`elimInv_acc_with_close`](https://github.com/ISTA-PLV/iris-lean/blob/1719ff7/Iris/Iris/ProofMode/Instances.lean#L961-L981)), which means `ProofMode/Instances.lean` has to import the tactics. So the dependency chain is basically `ProofMode/Classes.lean > IPM tactics > ProofMode/Instances.lean`. At some point, some code for IPM tactics imported the instances directly, which would cause circular dependency, and a previous attempt to resolve that resulted in `AsEmpValid` instances in `ProofMode/Instances.lean` not being recognised and `istart` failing in some random file.

It might be helpful to have something analogous to [`Init.lean`](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Init.lean) in Mathlib, which is imported by virtually all files for basic functionalities, which possibly makes it easier to maintain module dependencies, especially when there are modules for the proof mode depending on files under `BI` and vice versa.

alvinylt:
Regarding the `istart` problem, I suspect that previously the dependency chain was:
- `ProofMode/Classes.lean` > `ProofMode/Instances.lean` > IPM tactics > some file that uses IPM tactics

During the previous attempt to resolve the circular dependency, it became:
- `ProofMode/Classes.lean` > IPM tactics > `ProofMode/Instances.lean` (instances in `ProofMode/Instances.lean` proven using IPM tactics were at the end of the file)
- `ProofMode/Classes.lean` > IPM tactics > some file that uses IPM tactics (`AsEmpValid` instances not found)

And now there is no issue with the file as `ProofMode/Instances.lean` is imported by *some* upstream module.

markusdemedeiros:
Thanks for fixing this! Can confirm it's been an issue in downstream repos too. Creating an `Init.lean` sounds like a great idea.

markusdemedeiros:
Regarding the circular dependency, maybe we should run `lake shake` on the repo? I've never actually used it, but it might be good to see if there are any unused dependencies here as well.

alvinylt:
> maybe we should run `lake shake` on the repo?

That's exactly the tool I've wanted! With those suggestions we are already able to remove lots of redundant dependencies.

markusdemedeiros:
Awesome. Just to check: is this good to merge now?

alvinylt:
> Awesome. Just to check: is this good to merge now?

If this looks good to you feel free to merge.

I have experimented with `lake shake`, which provides the `--fix` option to apply the suggestions automatically, but that would cause immediately problems with the build. So I reckon dependencies are something to look into as a separate issue.

PR #555, opened 2026-07-30
fix: Wrong priorities of `AndIntoSep` instances

lzy0505:
## Description

Fix the incorrect behaviour of `icase` caused by the wrongly specified priorities of `AndIntoSep` instances. 

```
example [BI PROP] [BIAffine PROP] (P Q : PROP) :
  (□ P) ∧ <pers> Q ⊢ Q := by
  iintro H
  icases H with ⟨#_, HQ⟩
```
Before the fix, `icases` gave `∗HQ : <affine> <pers> Q`, which should be just `<pers> Q` following Iris Rocq.

## Checklist

## Generative AI Guidelines


Iris/Iris/ProofMode/Instances.lean
```diff
@@ -414,12 +414,13 @@ instance (priority := default + 10) fromSep_persistently [BI PROP] (P Q1 Q2 : PR
   from_sep := persistently_sep_mpr.trans (persistently_mono h.1)
 
 -- AndIntoSep
-@[ipm_class, rocq_alias AndIntoSep]
+@[ipm_backtrack, rocq_alias AndIntoSep]
```
MackieLoeffel: `ipm_backtrack` on a class does not make sense. It only makes sense on instances. Maybe you want to add the `ipm_backtrack` to `AndIntoSep.affine`?

MackieLoeffel (commented):
@alvinylt You also added some `ipm_backtrack` annotations in https://github.com/leanprover-community/iris-lean/pull/500 Are they solving the same problem or a different one?

Iris/Iris/ProofMode/Instances.lean
alvinylt: The fix in #500 has to do with `IntoAnd` rather than `AndIntoSep`. Here's an example, where ` □ Q1 ∗ □ Q2` matches both `intoAnd_sep` and `intoAnd_sep_affine`:

```lean4
example [BI PROP] (P Q1 Q2 : PROP) [Persistent Q1] [Persistent Q2] :
    ⊢ P -∗ (P -∗ □ Q1 ∗ □ Q2) -∗ P ∗ (P -∗ □ Q1 ∗ □ Q2) := by
  iintro HP HPQ
  ihave ⟨#HQ1, #HQ2⟩ : □ Q1 ∗ □ Q2 $$ [HP HPQ]
  · iapply HPQ $$ HP
  · isplitl [HP] <;> iassumption
```

Trace:

```
[Meta.synthInstance] ✅️ IPM: ProofMode.IntoAnd true iprop(□ Q1 ∗ □ Q2) iprop(□ Q1) iprop(□ Q2) ▼
  [] ✅️ new goal ProofMode.IntoAnd true iprop(□ Q1 ∗ □ Q2) ?m.98
        ?m.99 => ProofMode.IntoAnd true iprop(□ Q1 ∗ □ Q2) iprop(□ Q1) iprop(□ Q2) ▼
    [tactics] [] 
    [instances] #[@ProofMode.intoAnd_sep_affine, @ProofMode.intoAnd_sep] 
    [] ❌️ apply @ProofMode.intoAnd_sep to ProofMode.IntoAnd true iprop(□ Q1 ∗ □ Q2) ?m.98 ?m.99 ▶
    [] ✅️ apply @ProofMode.intoAnd_sep_affine to ProofMode.IntoAnd true iprop(□ Q1 ∗ □ Q2) ?m.98 ?m.99 ▶
  [] result ProofMode.intoAnd_sep_affine true iprop(□ Q1) iprop(□ Q2) 
  ```

So I think this PR solves a different problem.

lzy0505: Indeed. It seems like we don't even need to make it an `ipm_class`.

MackieLoeffel: I think AndIntoSep should remain an `ipm_class` since it uses `FromAffinely`, which is also an `ipm_class`. I've updated the PR.

MackieLoeffel:
Thanks for the PR!

PR #524, opened 2026-07-18
feat: lazy coin example

ayhon:
## Description
Adds the `lazy_coin.v` example from Iris Rocq as `LazyCoin.lean`. Because `lazy_coin.v` depends on `nondet_bool.v`, this PR includes `NondetBool.lean` as well.

## Checklist
* [ ] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [ ] I have added my name to the `authors` section of any appropriate files

PR #559, opened 2026-07-31
fix: `iinv` fails on Heaplang wp goals

lzy0505:
## Description
Make `Expr`, `Val`, `State` and `Obs` of `IrisGS_gen` `outParam` so that `iinv` works on Heaplang wp goals.

Before the fix, `iinv` failed with "ElimInv type class synthesis failed", because it failed at solving `IrisGS_gen` with input mvars. 

I couldn't find a better solution. What are your thoughts? @MackieLoeffel @alvinylt 

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/ProgramLogic/WeakestPre.lean
+++ b/Iris/Iris/ProgramLogic/WeakestPre.lean
@@ -42,8 +42,8 @@ class StateInterp (State : Type _) (Obs : outParam $ Type _) (GF : BundledGFunct
 export StateInterp (stateInterp)
 
 @[rocq_alias irisGS_gen]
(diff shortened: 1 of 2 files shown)


MackieLoeffel:
It is very funny that you are posting this now since I was looking into exactly the same problem this morning as part of a pass over the lazy coin PR :D 

I think the problem is that there is a diamond in the typeclass instances between HeapLangGS and IrisGS. I solved the problem in https://github.com/leanprover-community/iris-lean/pull/524/changes/47b494eb48ffc91493ffe11249d4f130b86f001c by removing this diamond.

lzy0505:
Nice! You fix is better. Closing

PR #561, opened 2026-07-31
feat: Port `IntoWand'` and instances

lzy0505:
## Description

Two of the `rocq_ignore`d `IntoWand'` instances seem necessary for `iapply` under update modalities.
Why were they ignored?

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -78,55 +78,46 @@ class FromWand {PROP} [BI PROP] (P : PROP) (io : InOut)
   from_wand : (Q1 -∗ Q2) ⊢ P
 export FromWand (from_wand)
 
-/--
-[WandMode.Side] names which of the two slots of a two-sided class is the input,
-in the modings that have one: `argument` or `result`.
--/
 inductive WandMode.Side where
   | argument
   | result
 
 /--
-[WandMode] describes the modings of a two-sided class such as `IntoWand`, by
-recording whether each of the argument and result slots is an input or an output.
+[WandMode] describes the modings of a two-sided class `IntoWand`, by recording whether each of the
+argument and result slots is an input or an output.
 
-`unknown` leaves both slots as outputs, and corresponds to Rocq's `IntoWand`
-(Hint Mode `! - -`). `balancing s` makes the slot `s` an input, and corresponds to
-Rocq's `IntoWand'` (Hint Modes `! ! -` and `! - !`).
+`unknown` leaves both slots as outputs, and corresponds to Rocq's `IntoWand` (Hint Mode `- -`).
+`matching s` makes the slot `s` an input, and corresponds to Rocq's `IntoWand'`
+(Hint Modes `! -` and `- !`).
+
+The priority of `matching` mode instances (such as `intoWand_bupd_args`) must stay below that of
+every instance with `unknown` (such as `intoWand_bupd`). This mirrors the priority `100` on
+Rocq's `into_wand_wand'`.
 -/
 inductive WandMode where
   | unknown
-  | balancing (s : WandMode.Side)
+  | matching (s : WandMode.Side)
 
 meta section
 
 /-- Whether the argument slot of a class at mode `m` is an input or an output. -/
 @[reducible]
-def WandMode.Side.argIO : WandMode.Side → InOut
-  | .argument => .in
-  | .result => .out
-
-/-- Whether the result slot of a class at mode `m` is an input or an output. -/
-@[reducible]
-def WandMode.Side.resIO : WandMode.Side → InOut
-  | .argument => .out
-  | .result => .in
-
-/-- Whether the argument slot of a class at mode `m` is an input or an output. -/
-@[reducible]
 def WandMode.argIO : WandMode → InOut
   | .unknown => .out
-  | .balancing s => s.argIO
+  | .matching .argument => .in
+  | .matching .result => .out
+
 
 /-- Whether the result slot of a class at mode `m` is an input or an output. -/
 @[reducible]
 def WandMode.resIO : WandMode → InOut
   | .unknown => .out
-  | .balancing s => s.resIO
+  | .matching .argument => .out
+  | .matching .result => .in
 
 end
 
-#rocq_ignore IntoWand' "the `WandMode` parameter of `IntoWand` subsumes it"
+#rocq_ignore IntoWand' "the `matching` mode of `IntoWand` subsumes it"
 
 @[ipm_class, rocq_alias IntoWand]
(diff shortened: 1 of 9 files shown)


MackieLoeffel:
These instances were ignored because the use `IntoWand'` instead of `IntoWand` in Rocq (notice the prime). I never understood what the purpose of the prime version is, so I just ignored them with the hope that at some point we will notice what they are useful for. Maybe this point is now. Do you know what the difference between `IntoWand'` and `IntoWand` in Rocq is and how we should map it to Lean?

lzy0505:
The only difference is the hint mode. 
In Rocq,`IntoWand` has `! - -` and `IntoWand'` has `! - !` and `! ! -` for the three arguments `R P Q`.

I think the main reason is to avoid divergence.

Consider the following instance that uses only `IntoWand` (Rocq uses the prime version on conclusion):
```
  Global Instance into_wand_bupd_args_BAD p q R P Q :
    IntoWand p false R P Q → IntoWand p q R (|==> P) (|==> Q).
```

It would diverge on query `IntoWand q false R ?P ?Q`.

I've updated the PR with a more systematic fix, and ported more `IntoWand'` instances.

Iris/Iris/Tests/Instances.lean
```diff
@@ -73,17 +73,18 @@ section mvars
 variable [BI PROP] (P1 P2 : Nat → PROP)
 
 /- Test creation of mvars -/
-/-- info: solution: IntoWand false false iprop(∀ x, P1 x -∗ P2 x) InOut.out (P1 ?m.24) InOut.out
-  (P2 ?m.24), new goals: [?m.24: Nat] -/
-#guard_msgs in #ipm_synth (IntoWand false false iprop(∀ a, P1 a -∗ P2 a) .out _ .out _)
+/-- info: solution: IntoWand false false iprop(∀ x, P1 x -∗ P2 x) WandMode.unknown (P1 ?m.23) (P2 ?m.23), new goals: [?m.23: Nat] -/
+#guard_msgs in #ipm_synth (IntoWand false false iprop(∀ a, P1 a -∗ P2 a) .unknown _ _)
```
MackieLoeffel, line 78: This should use `set_option pp.mvars false in` to avoid mvar numbers in guard_msgs.

Iris/Iris/Tests/Tactics.lean
```diff
@@ -1497,6 +1540,23 @@ example [BI PROP] (P Q : PROP) :
...
+-- `Q` is not a wand, so no `IntoWand` instance applies. This fails immediately instead of looping by
+-- wrapping the two output slots in ever more modalities.
+/-- error: ispecialize: Q is not a wand
+---
+trace: [Meta.synthInstance] ❌️ IPM: ProofMode.IntoWand false false Q ProofMode.WandMode.unknown ?m.36 ?m.37
+  [Meta.synthInstance] ❌️ IPM: new goal ProofMode.IntoWand false false Q ProofMode.WandMode.unknown ?m.36
+        ?m.37 => ProofMode.IntoWand false false Q ProofMode.WandMode.unknown ?m.36 ?m.37
+    [Meta.synthInstance.tactics] []
+    [Meta.synthInstance.instances] #[]
+  [Meta.synthInstance] result <not-available>
```
MackieLoeffel: What is the point of having this trace? I am a bit worried that this might make this test quite brittle.

Iris/Iris/ProofMode/InstancesUpdates.lean
```diff
@@ -33,25 +33,32 @@ instance fromPure_bupd (a : Bool) (P : PROP) (φ : Prop)
...
 
-set_option synthInstance.checkSynthOrder false in
+/-- Balances the argument and the result of `R` against the goal's modality.
+Since `R` is left untouched, this only terminates when one of the two slots is an
+input, so it is stated as `balancing _` and never at `unknown`. The `Side` is left a
+variable, so this single instance covers both `argument` and `result`.
+
+The priority must stay below that of every instance which strips structure off
+`R` (such as `intoWand_bupd`), so that this is only reached once `R` has bottomed
+out. This mirrors the priority `100` on Rocq's `into_wand_wand'` bridge. -/
```
MackieLoeffel: This comment is great, but it is a bit hard to find and we don't want to repeat it above all `_args` instances. Maybe we should move it to where `.balancing` is defined?

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -78,12 +78,60 @@ class FromWand {PROP} [BI PROP] (P : PROP) (io : InOut)
...
+
+/-- Whether the argument slot of a class at mode `m` is an input or an output. -/
+@[reducible]
+def WandMode.Side.argIO : WandMode.Side → InOut
+  | .argument => .in
+  | .result => .out
+
+/-- Whether the result slot of a class at mode `m` is an input or an output. -/
+@[reducible]
+def WandMode.Side.resIO : WandMode.Side → InOut
```
MackieLoeffel: It seems like `WandMode.Side.argIO` and `WandMode.Side.resIO` are only used in `WandMode.argIO` and `WandMode.resIO`. So maybe we can just inline them there?

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -78,12 +78,60 @@ class FromWand {PROP} [BI PROP] (P : PROP) (io : InOut)
...
+[WandMode] describes the modings of a two-sided class such as `IntoWand`, by
+recording whether each of the argument and result slots is an input or an output.
+
+`unknown` leaves both slots as outputs, and corresponds to Rocq's `IntoWand`
+(Hint Mode `! - -`). `balancing s` makes the slot `s` an input, and corresponds to
+Rocq's `IntoWand'` (Hint Modes `! ! -` and `! - !`).
+-/
+inductive WandMode where
+  | unknown
+  | balancing (s : WandMode.Side)
```
MackieLoeffel: Maybe instead of `balancing` use `matching`? I would say that the wand mode determines which side of the wand should be matched (if any).

MackieLoeffel (commented):
Great work figuring this out and I like your solution using @alvinylt 's dynamic hint mode feature! I left some minor style comments.

Iris/Iris/Tests/Tactics.lean
lzy0505: This is to check that the modings of `IntoWand` work - the typeclass search fails quickly instead of looping. I don't know how to check it without showing the instance causing the loop is not available in the trace. 

I've updated the comments and make `guard_msg` only match on the substring. Is it better now?

MackieLoeffel: Yes, this is better. :+1:

MackieLoeffel:
Thanks for fixing this!

PR #563, opened 2026-07-31
feat: make MaxNat distinct from Nat

### LINKED ISSUE #562: `instAddMaxNat` is leaking out of Iris Lean
To reproduce, open a file in the root of this repository and type
```lean4
module

public meta import Iris.Algebra.Lib.MonoNat

namespace Iris

#guard 1 = 1 + 1
```

## Expected behavior
The guard fails, since `1 + 1 = 2` (or so says [Principia Mathematica](https://en.wikipedia.org/wiki/Principia_Mathematica))

## Actual behavior
Since `MaxNat = Nat` and for `MaxNat` addition is `max`, the guard succeeds.

ayhon:
## Description
This PR makes `MaxNat` a structure carrying a `Nat` value, instead of an abbreviation. This prevents the `Add` instance for `MaxNat` from ever being used as a `Add Nat` instance, like it was the case for #562.

Fixes #562

## Checklist


MackieLoeffel:
It makes sense that `MaxNat` is a separate type from `Nat` so I am in favor of this PR.
[This is also how Rocq does it.](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/numbers.v#L50)
But as mentioned in #562 , I am not sure why we want to have an `Add` Instance that is implemented as `max`.

ayhon:
> It makes sense that `MaxNat` is a separate type from `Nat` so I am in favor of this PR.
> [This is also how Rocq does it.](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/algebra/numbers.v#L50)
> But as mentioned in #562 , I am not sure why we want to have an `Add` Instance that is implemented as `max`.

We could also refactor the code to remove the Add instance. I feel like this is actually just a detail in the implementation, the instance seems to only be needed for the UCMRA definition

MackieLoeffel:
I think we should make MaxNat a custom structure instead of an abbrev in any case, like is already the case in Rocq. Otherwise there will be a lot of type confusion. If we can additionally also make the `Add` instance local, this would probably also be good.

markusdemedeiros:
I think that making `MaxNat` a separate type is also probably the way to go. Having it be an instance on `Nat` (which, yes, should have been local) is nice for avoiding extra indirections and improving automation, but I don't think that applies here.

markusdemedeiros:
It sounds like we're in agreement so I'm going to merge this

markusdemedeiros:
I'm introducing a similar construction to also knock out `maxZ`, so if we tweak this we should change it here too.

PR #564, opened 2026-08-01
fix: make `instMaxNatAdd` a `local` instance

### LINKED ISSUE #562: `instAddMaxNat` is leaking out of Iris Lean

ayhon:
## Description
Makes `instMaxNatAdd` a `local` instance.

Fixes #562.
Alternative to #563.

## Checklist

### DIFF
--- a/Iris/Iris/Algebra/Lib/MonoNat.lean
+++ b/Iris/Iris/Algebra/Lib/MonoNat.lean
@@ -18,7 +18,7 @@ section MaxNat
 
 abbrev MaxNat := Nat
 
-scoped instance : Add MaxNat := ⟨max⟩
+local instance : Add MaxNat := ⟨max⟩
 scoped instance : Associative (Add.add (α := MaxNat)) where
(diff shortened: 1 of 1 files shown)


MackieLoeffel:
Won't this break things for users of `MaxNat` that expect to have this instance?

ayhon:
> Won't this break things for users of `MaxNat` that expect to have this instance?

As far as I could tell, this operation is only ever used to define instances. Elsewhere, the operation is referred to as any other `op`, since Add is only needed to make the UCMRA instance.

PR #565, opened 2026-08-03
feat: update to v4.32.2

srghma:
## Description
just updates toolchain

## Checklist

### DIFF
--- a/Iris/lakefile.toml
+++ b/Iris/lakefile.toml
@@ -4,12 +4,12 @@ defaultTargets = ["Iris", "IrisTest"]
 [[require]]
 name = "Qq"
 scope = "leanprover-community"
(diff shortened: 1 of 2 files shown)


markusdemedeiros:
The project is already on 4.32.2. Your PR only updates the Qq dependency, but keeping Qq on 4.32.0 is intentional (they did not create a 4.32.2 release).

srghma:
in https://reservoir.lean-lang.org/@leanprover-community/Qq I see they did create a 4.32.2 release

markusdemedeiros:
I don't think that's a release--that's just reservoir saying that the latest commit also builds on these newer versions (which is expected since Iris-Lean still compiles). It is better to keep dependencies pinned to real releases.

PR #567, opened 2026-08-03
fix: correct parallel composition syntax and add test for `par`

kdvkrs:
## Description
As @suhr pointed out in [this Zulip message](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Porting.20iris-tutorial/near/613886178), substitution currently does not descend into `‖` expressions because the notation wraps its thunks in value lambdas.

This PR changes `hl($e1 ‖ $e2)` to use expression lambdas (c.f. [Iris-Rocq’s expression-scope `|||` notation)](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris_heap_lang/lib/par.v?ref_type=heads#L13). `wp_par` retains the value-lambda form, matching its `%V` use in Iris-Rocq.

The PR also adds @suhr’s example as a regression test.

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/Tests/HeapLang/Par.lean
+++ b/Iris/Iris/Tests/HeapLang/Par.lean
@@ -12,11 +12,18 @@ namespace Iris.Tests.HeapLang.Par
 
 open Iris.HeapLang BI Iris ProgramLogic Spawn Iris.HeapLang.Par
 
+-- Regression test for
+-- https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Porting.20iris-tutorial/near/613886178
+-- testing substitution into `par`
+section
+-- `l` is deliberately a free HeapLang variable so we can test that substitution works as intended
 set_option linter.heapLang.freeVars false
 
 example (v : Val) :
     Exp.substStr "l" v hl((l ← #21) ‖ (l ← #2))
       = hl(&par (λ _, v(&v) ← #21) (λ _, v(&v) ← #2)) := rfl
+
+end
 
 def par_client : Exp := hl%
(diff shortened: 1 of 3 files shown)


Iris/Iris/Tests/HeapLang/Par.lean
```diff
@@ -0,0 +1,55 @@
...
+module
+
+public import Iris.HeapLang.Lib.Par
+
+@[expose] public section
+namespace Iris.Tests.HeapLang.Par
+
+open Iris.HeapLang BI Iris ProgramLogic Spawn Iris.HeapLang.Par
+
+set_option linter.heapLang.freeVars false
```
MackieLoeffel, line 20: Why is this necessary? If it is necessary just for the `example` below, it is better to use `set_option ... in`.

Iris/Iris/Tests/HeapLang/Par.lean
```diff
@@ -0,0 +1,55 @@
...
+
+example (v : Val) :
```
MackieLoeffel, line 22: Please add a comment to refer to the bug report that this is testing.

MackieLoeffel (commented):
Thanks for the PR!

Iris/Iris/Tests/HeapLang/Par.lean
kdvkrs: This is only for the example since `l` is unbound and Lean would warn without turning the option. `set option ... in` doesn't seem to suppress the warning. I've put it into a section instead.

MackieLoeffel:
Great, thanks for the PR! Merging.

PR #568, opened 2026-08-04
Make Texan triples persistent

suhr:
## Description

Make Texan triples persistent because they really should be.

I have noticed that Rocq Iris defines triples with `□` in `bi_scope`, but without it in `stdpp_scope`. But what is the `stdpp_scope`?

## Checklist


MackieLoeffel:
`stdpp_scope` in Rocq is a top-level scope that is always open. In Lean, this corresponds to notations that are directly in the `term` syntax category, not below an `iprop(...)`. It would be nice to have a `term` notation for Texan triples, then we could omit the `⊢` before it. For this notation, we don't need the box since top-level entailments can always be reused.

suhr:
I think that a `term` notation could be added in a next PR.

ayhon:
I implemented the `term` notation in [this PR](https://github.com/suhr/iris-lean/pull/7) ([branch](https://github.com/ayhon/iris-lean/tree/fix/texan-triple-persistent)). @suhr, feel free to merge it into this PR if you'd like. 

The issue with the term notation is that it requires one to specify the `GF` from the `IProp` somewhere in the precondition or postcondition, since the Texan triple syntax doesn't provide a way to specify the underlying `PROP` the same way entailment does with `|-@{...}`

suhr:
I rebased the last two commits from your branch onto my branch.

Iris/Iris/Tests/HeapLang/Par.lean
```diff
@@ -36,7 +36,7 @@ example {hlc} {GF : BundledGFunctors} [HeapLangGS hlc GF] [SpawnG GF] :
     ⊢ {{ True }} hl(&par_client)
```
MackieLoeffel: This should also use the term notation instead of the iprop notation.

Iris/Iris/BI/WeakestPre.lean
```diff
@@ -104,8 +104,26 @@ meta def wpMacro : Lean.Macro := fun stx => do
...
+              `(iprop(∀ $xs*, $Q:term -∗ Φ $pat))
+            | none => `($Q:term -∗ Φ $pat)
+    `(iprop(□ ∀ Φ, $P -∗ ▷ $k -∗ (WP $wpExpr {{ Φ }})))
+  | _ => Lean.Macro.throwUnsupported
+
+@[macro texanTriple]
+meta def wpTexanTripleTerm : Lean.Macro
   | `({{ $P:term }} $wpExpr {{ $[$[$xs]* ,]? RET $pat ; $Q:term }}) => do
 
     let transform (xs : Array (TSyntax [`Lean.binderIdent, `Lean.Parser.Term.bracketedBinder])) : MacroM <| TSyntaxArray [`ident, `Lean.Parser.Term.hole, `Lean.Parser.Term.bracketedBinder] :=
```
MackieLoeffel: Instead of duplicating this code, can we just desugar this to `⊢ {{ ... }} ... {{ ... }}` and then continue elaboration there?

ayhon: You mean have two separate macros
 - `{{ .. }} .. {{ .. }}` => `⊢ term({{ .. }} .. {{ .. }})` 
 - `term({{ ... }} ... {{ ... }})` => `∀ Φ, $P -∗ ▷ $k -∗ (WP $wpExpr {{ Φ }})`, this one without a `□`
 
 ?

ayhon: I don't know if `term(iprop(_))` => `_` would trigger beforehand, maybe we need to translate the first one into `BI.Entails iprop(emp) term({{ .. }} .. {{ .. }})` to prevent this.

MackieLoeffel: No, my idea idea was
- {{ .. }} .. {{ .. }} => ⊢ iprop({{ .. }} .. {{ .. }})
- iprop({{ ... }} ... {{ ... }}) => □ ∀ Φ, $P -∗ ▷ $k -∗ (WP $wpExpr {{ Φ }})

But this adds the `□` also for the `term` notation. Not sure if there is a way to avoid this.

MackieLoeffel: Alternatively, we can just extract the shared code to a function that is used by both macros.

suhr: I think, the implementation could be refactored in a next PR.

MackieLoeffel: I moved out the code to a function.

MackieLoeffel:
Thanks for the PR! It looks good to me now.

PR #578, opened 2026-08-08
feat: Bring Several more Algebra files to 100% 

markusdemedeiros:
## Description
Brings the following files to 100% ported

- Agree
- Auth
- BigOp
- COFESolver
- ~~DFrac~~
- DFracAgree
- Excl
- ExclAuth
- Frac
- FracAuth
- Functions
- IsOp
- LeibnizSet
- LocalUpdates
- MonoNat
- Monoid
- StepIndex
- UFrac
- UFracAuth
- View

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/Algebra/DFrac.lean
+++ b/Iris/Iris/Algebra/DFrac.lean
@@ -44,8 +44,6 @@ open DFrac OFE.Discrete IsOp
 @[rocq_alias dfrac_inhabited]
 instance : Inhabited DFrac := ⟨discard⟩
 
-#rocq_ignore dfrac_countable "std++ `Countable`; Lean has no `Pos.Countable Qp` and does not need one"
-
 def valid : DFrac → Prop
   | .own f        => f.val ≤ 1
   | .discard      => True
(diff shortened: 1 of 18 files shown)


Iris/Iris/Algebra/DFrac.lean
```diff
@@ -41,6 +44,8 @@ open DFrac OFE.Discrete IsOp
 @[rocq_alias dfrac_inhabited]
 instance : Inhabited DFrac := ⟨discard⟩
 
+#rocq_ignore dfrac_countable "std++ `Countable`; Lean has no `Pos.Countable Qp` and does not need one"
```
markusdemedeiros: Inaccurate

Iris/Iris/Algebra/Updates.lean
```diff
@@ -69,6 +69,10 @@ instance [CMRA α] : Trans Update Update Update (α := α) where
 instance [CMRA α] : Trans Update UpdateP UpdateP (α := α) where
   trans := Update.transP
 
+#rocq_ignore cmra_update_preorder "Use Update.id and Update.trans"
```
markusdemedeiros: Is there really no preorder class in core?

markusdemedeiros (commented):
Double checked this manually. The AI incorrectly `rocq_ignore`'d some things, saying they were synthesizable even when they were not (fixed the synthesis in these cases, glad something good came of this!) Also, I will kick the can on the countability instance in DFrac. It needs Qp to be countable, which is an annoying proof and I'm not sure yet if we should change to plain rationals instead. Will think about this some other time.

lzy0505 (approved):
Nice!

PR #589, opened 2026-08-09
chore: Lean and CI scripts to check for missing module imports

alvinylt:
## Description

A follow-up on #552.

When new modules are introduced into the Mathlib repository, `lake exe mk_all` is used for updating `Mathlib.lean`, the entry point for building all modules. It will be nice to have something similar so that new modules are not missed by [`Iris.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris.lean) in the future.

This PR introduces a script written in Lean to check for missing imports. Basically it tracks the dependency chain, traverses the directories and compares the difference between the two. One can run the script using `lake exe check-imports <LibraryName>`. For example, by running `lake exe --subdirs check-imports Iris` with the current state of the `master` branch, we get:

```
check-imports: 2 file(s) under Iris are never imported (directly or transitively) from Iris.lean:
  Iris.ProgramLogic
  Iris.ProgramLogic.EctxLifting
check-imports: 1 file(s) under Iris/Algebra are never imported (directly or transitively) from Iris/Algebra.lean:
  Iris.Algebra.Numbers
check-imports: 1 file(s) under Iris/BI are never imported (directly or transitively) from Iris/BI.lean:
  Iris.BI.Algebra
check-imports: all 7 modules under Iris/Examples are imported from Iris/Examples.lean.
check-imports: all 9 modules under Iris/HeapLang are imported from Iris/HeapLang.lean.
check-imports: all 10 modules under Iris/Instances are imported from Iris/Instances.lean.
check-imports: all 11 modules under Iris/ProgramLogic are imported from Iris/ProgramLogic.lean.
check-imports: all 54 modules under Iris/ProofMode are imported from Iris/ProofMode.lean.
check-imports: all 29 modules under Iris/Std are imported from Iris/Std.lean.
check-imports: all 14 modules under Iris/Tests are imported from Iris/Tests.lean.
```

The flag `--subdirs` enables checking not only `Iris.lean` but also entry-point modules directly under the `Iris/` directory (e.g. `Algebra.lean`, `BI.lean`). I assume the repository will continue to be structured in this way, but the script can be generalised/modified if we would like to do this recursively.

The CI script is updated to run this once `Iris/` and `IrisMath/` so that any missing modules are flagged.

This is less powerful than `lake exe mk_all` in Mathlib, as the latter is able to modify `Mathlib.lean` directly, but our script is also much simpler and avoids additional dependencies as we are only using built-in functions provided by Lean.

The files [`Std/DumpPortingData.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/Std/DumpPortingData.lean) and [`Std/Porting.lean`](https://github.com/leanprover-community/iris-lean/blob/master/Iris/Iris/ProofMode/Porting.lean) are not modules and cannot be imported. At the moment, I have hardcoded their exclusion in the script. Not sure what the best solution to this is, and would like to hear any suggestions.

### Other Changes

- Adds those missing imports, as shown above.
- The IPM tactic `iintro` is used only once within [`BI/Algebra.lean`](https://github.com/ISTA-PLV/iris-lean/blob/37f53e0/Iris/Iris/BI/Algebra.lean#L147). By importing the proof mode module here we easily get into circular dependencies between `BI` and `ProofMode`, but this is easily avoidable.

## Checklist


IrisMath/lakefile.toml
```diff
@@ -8,7 +8,7 @@ pp.unicode.fun = true
 relaxedAutoImplicit = true
 weak.linter.mathlibStandardSet = true
 maxSynthPendingDepth = 3
-linter.style.header = false
+weak.linter.style.header = false
```
MackieLoeffel, line 11: What does this do?

scripts/CheckImports.lean
```diff
@@ -0,0 +1,142 @@
...
+
+open System (FilePath)
+
+abbrev leanSuffix := ".lean"
+
+/-- Convert path name to module name (e.g. `Iris/BI/BIBase.lean` to `Iris.BI.BIBase`). -/
+private def pathToModule (p : FilePath) : String :=
+  let s := p.toString
+  let s := if s.endsWith leanSuffix then s.dropEnd leanSuffix.length else s
+  (s.replace "\\" "/").replace "/" "."
```
MackieLoeffel: This seems a bit hacky. Is there no builtin function for this?

scripts/CheckImports.lean
```diff
@@ -0,0 +1,142 @@
...
+  -- Strip leading modifiers
+  let rec dropMods : List String → List String
+    | "public" :: rest => dropMods rest
+    | "meta" :: rest => dropMods rest
+    | "private" :: rest => dropMods rest
+    | rest => rest
+  -- Remove the leading `import` keyword
+  match dropMods toks with
+  | "import" :: modName :: _ => some modName
+  | _ => none
```
MackieLoeffel: I am wondering if there is a version of this, which does not require us to manually parse Lean files.

MackieLoeffel (commented):
Thanks for this PR! This is a great idea. The implementation looks a bit hacky to me, but if this is the easiest version to implement, I am also fine with this.

alvinylt:
The script is rewritten to reuse functions provided by Lean's core libraries and to be more generic. This avoids the manual parsing and kludges that involve string manipulation.

The functions `Lean.modToFilePath`, `Lean.forEachModuleInDir`, `Lean.findModuleWithExt` and `Lean.parseImports'` are available for conversion between file pathnames and Lean module names. They are directly used the implementation of `lake` and thus there is no need for our previous custom functions.

The most crucial part of the script are three assumptions regarding how modules under `Iris` and `IrisMath` are organised:
- Given `root ∈ { Iris, IrisMath }`, the directory `root/s_1/s_2/.../s_n` always comes with the "entry-point module" `root/s_1/s_2/.../s_(n-1)/s_n.lean`.
  - For example, `Iris/ProofMode.lean` is the entry-point module that imports all modules under the directory `Iris/ProofMode`, and `Iris/BI/BigOp.lean` is the entry-point module that imports all modules under the directory `Iris/BI/BigOp`.
  - If this requirement is satisfied on the top-level (`Iris`), then `Iris.lean` is guaranteed not to miss any modules under `Iris`.
- Exception 1: the files `Iris/ProofMode/Porting.lean` and `Iris/Std/DumpPortingData.lean` are not modules, so they are excluded.
   - These exceptions are listed in the array `excludedModules` within the script.
   - They are unconditionally exempted from the check.
- Exception 2: for `s ∈ { Algebra, BI, Instance, HeapLang }`, `Iris/s.lean` does not import modules under the directory `Iris/s/Lib`. (Otherwise there would be some circular dependency.)
  - These exceptions are listed in the array `detachedDirs` within the script.
  - For every `Iris.s.Lib` in `detachedDirs`, the script accepts that the immediate parent (`Iris/s.lean`) does not import `Iris.s.Lib` or its submodules. However, the exemption does not apply to transitive parents. That is, `Iris.lean` still has to import all those submodules under `Lib`.

The script is now generic enough to do the recursive check, so there is no need for the special `--subdirs` flag.

This is as generic as we can get while fitting into the current structure of the codebase and also making it easy to change the exceptions if necessary. How does this sound?

Commit: [4851daa](https://github.com/leanprover-community/iris-lean/pull/589/commits/4851daaa4dc259665e577ed321bb93d9767469a9)

scripts/CheckImports.lean
alvinylt: See comment below.

scripts/CheckImports.lean
alvinylt: See comment below.

IrisMath/lakefile.toml
alvinylt: With `weak.linter.style.header = false`, the linter config is set as `false` if the option `linter.style.header` exists, or otherwise ignored. In contrast, `linter.style.header = false` in `lakefile.toml` would result in an error if the option does not exist (i.e. there is no linter enabled).

This does not matter for building `IrisMath` as the code style linter is available as a Mathlib dependency. The new script `CheckImport.lean` does not import Mathlib itself, so it would throw an error about the invalid config.

An existing line above it (`weak.linter.mathlibStandardSet = true`) is similar.

alvinylt:
Check out the [failed CI run](https://github.com/leanprover-community/iris-lean/actions/runs/31429771377/job/93590038681?pr=589) in commit [8b1c423](https://github.com/leanprover-community/iris-lean/pull/589/commits/8b1c42369e5034fdb4868fda496c6a97197f354c) above, where some of the imports are removed.

```
check-imports: 6 file(s) under Iris are never imported (directly or transitively) from ./Iris.lean:
  Iris.Algebra.DynReservationMap
  Iris.BI.BigOp.BigAndMap
  Iris.BI.Lib.FixpointBanach
  Iris.BI.Lib.Laterable
  Iris.ProgramLogic
  Iris.ProgramLogic.EctxLifting
check-imports: 1 file(s) under Iris.Algebra are never imported (directly or transitively) from ./Iris/Algebra.lean:
  Iris.Algebra.DynReservationMap
check-imports: 1 file(s) under Iris.BI are never imported (directly or transitively) from ./Iris/BI.lean:
  Iris.BI.BigOp.BigAndMap
check-imports: 1 file(s) under Iris.BI.BigOp are never imported (directly or transitively) from ./Iris/BI/BigOp.lean:
  Iris.BI.BigOp.BigAndMap
check-imports: 3 file(s) under Iris.BI.Lib are never imported (directly or transitively) from ./Iris/BI/Lib.lean:
  Iris.BI.Lib.FixpointBanach
  Iris.BI.Lib.GenHeap
  Iris.BI.Lib.Laterable
```

MackieLoeffel (approved):
I think this looks great! @markusdemedeiros @lzy0505 Any thoughts on this? Otherwise, I think this can be merged.

lzy0505 (approved):
Great work!

PR #591, opened 2026-08-09
feat: improve proof state display within IPM

alvinylt:
## Description

Addresses the Zulip discussion thread [#iris-lean > Porting iris-tutorial @ 💬](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/Porting.20iris-tutorial/near/615480790).

This involves some improvements to the IPM goal delaborator so that the type signature of an item is shown in a pop-up message upon hovering instead of having the entire entailment as a single item.

Pop-up message showing the type signature of an item in the Iris Proof Mode

The proof mode will be even more usable when more syntax definitions are annotated with docstrings.

## Checklist

### DIFF
--- a/Iris/Iris/ProofMode/Display.lean
+++ b/Iris/Iris/ProofMode/Display.lean
@@ -7,7 +7,6 @@ module
 
 public meta import Iris.BI.Notation
 public meta import Iris.ProofMode.Expr
-
 public meta import Lean.PrettyPrinter.Delaborator
 
 public meta section
--- a/Iris/Iris/ProofMode/Display.lean
+++ b/Iris/Iris/ProofMode/Display.lean
@@ -39,12 +39,20 @@ def withHypType {α} [Inhabited α] (persistent : Bool) (d : DelabM α) : DelabM
   let d := withMDataExpr <| withAppArg d
   if persistent then withNaryArg 2 d else d
 
+@[delab app.Iris.ProofMode.IrisHyp]
+def delabIrisHyp : Delab := withAppArg delab
+
 @[delab app.Iris.ProofMode.Entails']
 def delabIrisGoal : Delab := do
-  let some { hyps, .. } := parseIrisGoal? (← instantiateMVars (← getExpr)) | failure
-  let (_, hypStxs) ← withNaryArg 2 <| delabHypotheses hyps ({}, #[])
+  let some { hyps, goal, .. } := parseIrisGoal? (← instantiateMVars (← getExpr)) | failure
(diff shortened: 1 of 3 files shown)


MackieLoeffel (approved):
Thanks for the PR! This looks great. Is there some way to add a test for this?

alvinylt:
> Is there some way to add a test for this?

Not sure how though, the effects are visible interactively on InfoView. A buggy delaborator would result in the proof state being printed weirdly (it got into some infinite recursion whilst I was modifying the code), but all the proofs would still pass.

MackieLoeffel:
> Not sure how though

Maybe you can delaborate some expression and then look at / print the resulting [InfoTree](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/InfoTree/Types.html#Lean.Elab.InfoTree)? Not sure though how to do this and how verbose this is.

alvinylt:
> Maybe you can delaborate some expression and then look at / print the resulting [InfoTree](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/InfoTree/Types.html#Lean.Elab.InfoTree)?

Nice suggestion, I have added some tests that uses a custom elaborator to perform a depth first traversal of the items. The printed trace looks like:

```
⋆ H (binder) : <absorb> (P1 ∗ P2 ∗ <affine> (P3 ∗ P4)) -∗ Q
  ⋆ <absorb> (P1 ∗ P2 ∗ <affine> (P3 ∗ P4)) -∗ Q : PROP
    ⋆ (P1 ∗ P2 ∗ <affine> (P3 ∗ P4)) : PROP
      ⋆ P1 : PROP
      ⋆ P2 : PROP
      ⋆ (P3 ∗ P4) : PROP
        ⋆ P3 : PROP
        ⋆ P4 : PROP
```

This is actually quite useful as it reveals several issues in this example. The first issue is that the pop-up for `P3 ∗ P4` does not match the actual proposition.

A mismatch of the type signature in a pop-up

The second issue is that certain nodes such as `<affine> ...`, `<absorb> ...`, `P -∗ Q`, `P ∗ Q` are skipped. I reckon it is a bug with `unpackIProp` in `BI/Notation.lean`. Now trying to see if this can be fixed.

alvinylt:
Another issue is that the internal machinery for IPM pretty printing (e.g. `Entails'`, `IrisHyp`) is shown to the user. I've adjusted the delaborator to avoid this.

Before:

Internal machinery not concealed

After:

Internal machinery concealed

MackieLoeffel:
This looks great! Thanks!

PR #516, opened 2026-07-14
feat: port HeapLang pure step tactics

kdvkrs:
## Description
Port the wp_pure wrapper tactic family from `iris_heap_lang/proofmode.v`: `wp_lam`, `wp_let`, `wp_seq`, `wp_closure`, `wp_if`, `wp_if_true`, `wp_if_false`, `wp_proj`, `wp_inj`, `wp_pair`, `wp_unop`, `wp_binop`, `wp_op`, `wp_case`, and `wp_match`.

The Lean implementation follows the Rocq implementation with these tactics being mostly thin wrappers around wp_pure.

Raise the precedence of `wp_pure`'s focus argument to `hl_exp:10` mirroring `wp_bind`. At default precedence, e.g. `wp_pure <pat>; wp_lam` misparses the `;` as focus pattern instead of sequencing.

Add a test section per tactic covering at least one successful and one failing case.

Partially addresses #480 

*AI Disclaimer*: I am familiar with the Iris-Rocq HeapLang tactic implementation and used Claude Code to help navigate the codebase and assist with testcase boilerplate.

## Checklist
* [X] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [X] I have added my name to the `authors` section of any appropriate files

PR #592, opened 2026-08-09
refactor: `throwIPMError` for error messages by IPM tactics

alvinylt:
## Description

Addresses the discussions in https://github.com/leanprover-community/iris-lean/pull/516#issuecomment-4980039789.

There have been discussions about whether having the tactic name in the error message is actually useful at all. In my opinion, there is not much difference from the user's perspective, given that there is already error highlighting in the IDE, and the use of `withRef` enables the highlighting to be more precise (e.g. in `iinduction`, introduction patterns). Nonetheless, there are some minor benefits, such as avoiding repeated threading of tactic names in function calls (e.g. with `throwTacticEx`) and a more consistent error message format amongst HeapLang tactics and other tactics.

## Checklist

### DIFF
--- a/Iris/Iris/HeapLang/ProofMode.lean
+++ b/Iris/Iris/HeapLang/ProofMode.lean
@@ -180,14 +180,14 @@ public meta def ProofModeM.runTacticWp {α} (tacName : Name) (k : MVarId → WpG
   : TacticM α := do
   ProofModeM.runTactic tacName fun mvar {u, prop, bi, hyps, goal, ..} => do
     let .defEq _ ← isLevelDefEqQ u 0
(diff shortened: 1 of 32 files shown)


MackieLoeffel:
I like the idea of having a consistent function for throwing errors from the proof mode. This also means that in the future we can easily print more (or less) information, as we see fit. I would be happy to merge this version and if we want some other printing in the future, we can easily change it.
@lzy0505 Do you have an opinion on this?

MackieLoeffel (approved):

lzy0505 (approved):
This is a nice improvement and ready for merging!
We could see if we can make error messages more informative later.

MackieLoeffel:
Thanks for having a look. Merging.

Issue #145, opened 2026-02-03
Instance synthesis fails with a match expression


lzy0505:
```
/-- error: (match ?b with
  | true => ?Q
  | false => ?Q) ⊢
  ?Q is not an entailment -/
#guard_msgs in
example [BI PROP] (Q : PROP) : Q ⊢ Q := by
  iintro HQ
  have H: (∀ b (Q: PROP), (match b with | true => iprop(Q) | false => iprop(Q)) ⊢ Q) := by sorry
  iapply H
```

This example failed at `iapply H`. 
But the `iapply` works if I replace the match expression with `(if b == true then iprop(Q) else iprop(Q))`.

The problem seems to be at this line: 
https://github.com/leanprover-community/iris-lean/blob/1bed579a41e3306ff77d90367dfaae511eb8d59f/src/Iris/ProofMode/SynthInstance.lean#L86

Is this an upstream bug?

cc @MackieLoeffel

btj-h:
I bumped into this as well. I created an upstream issue.

MackieLoeffel:
Closed via #594

PR #594, opened 2026-08-10
fix: Do not throw isDefEqStuck in IPM TC synth

MackieLoeffel:
## Description
The IPM TC synth used the same defeq config as the standard Lean TC synth, which includes setting the flag for the isDefEqStuck exception. However, the IPM TC synth does not handle this exception, and thus the IPM TC synth can unexpectedly abort. This commit fixes this by disabling the isDefEqStuck exception.

Fixes [#iris-lean > iapply doesn't work with matches? @ 💬](https://leanprover.zulipchat.com/#narrow/channel/490604-iris-lean/topic/iapply.20doesn.27t.20work.20with.20matches.3F/near/615255205) and #145 

## Checklist


lzy0505 (approved):
Also fixes #145.

MackieLoeffel:
Indeed! I thought #145 was a separate issue, but great that this also resolves #145.

PR #596, opened 2026-08-10
chore: bump version to 4.33.0

markusdemedeiros:
This version introduces more strict namespacing rules, so this may be a bigger change. I think we should block on #576 which is also a relatively large pending change.


markusdemedeiros:
In fact, let's just wait a week and a half or so until the project is at 100%

markusdemedeiros:
Note to self: #625 has information about the gradual adoption of linters

Issue #211, opened 2026-04-07
Port bi/telescopes.v


lzy0505:
## Summary

Port [`bi/telescopes.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/bi/telescopes.v).

## Dependencies

**Rocq dependencies:**

- `bi/bi` (aggregation)

PR #610, opened 2026-08-11
ignore telescopes

ahuoguo:
## Description

This will deprecate #211 and #598

IIUC, the proofmode use of `bi.telescopes` is all replaced by lean metaprogramming, and similar tricks can be done when porting logically atomic triples.

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/BI/Porting.lean
+++ b/Iris/Iris/BI/Porting.lean
@@ -1,0 +1,9 @@
+/-
+Copyright (c) 2026. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Alex Bai
+-/
+import Iris.Std.RocqPorting
+
+#rocq_ignore_file bi "telescopes.v"
+  "Rocq-specific telescope infrastructure not needed in the Lean metaprogram"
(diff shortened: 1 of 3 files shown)


markusdemedeiros:
I don't really understand how telescopes are used, but yeah, if they really are Rocq-specific then I think it would be better to ignore with a message describing the Lean way. #598 was a purely blind porting effort 😅

MackieLoeffel:
I think telescopes might also be useful in Lean, as they are in Rocq. They are very useful when you have definitions that contain multiple binders and I know a few projects based on Iris that use them (often in the context of automation, e.g. Diaframe uses them and other automation). I think they might come in handy for Iris-Lean based projects as well at some point. So I would not ignore them. If we don't want to port them, we can just wait and see if someone needs them.

ahuoguo:
Thanks for your input! So far, I don't think I really need them for logatom triples.

I think the proofmode ignores are still valid? as documented here: https://github.com/leanprover-community/iris-lean/pull/511/changes#diff-5a2723cddf35ed52cb237eb9d7f363ad47dd49d9711e52394972ab514d9825eaR455-R456 Though I can see if someone ported the telescope architecture they might also want to port all of the relevant proofmode infra.

MackieLoeffel:
GatherEvarsEq is some specific part of the iframe machinery in Rocq, which does not appear in Lean. Its rocq_ignore just applies to the use of telescopes in iframe, not to telescopes in general.

PR #629, opened 2026-08-13
feat: Annotations for bupd_alt

markusdemedeiros:
## Description
There is one that needs a change to the annotations script to fix, but we should defer doing that until the very end. 

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/BI/Lib/BUpdPlain.lean
+++ b/Iris/Iris/BI/Lib/BUpdPlain.lean
@@ -1,11 +1,13 @@
+/-
+Copyright (c) 2026. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors:
+-/
 module
 
 public import Iris.Std
 public import Iris.Algebra.Updates
-public import Iris.ProofMode.Classes
-public import Iris.ProofMode.Tactics
-public import Iris.ProofMode.Display
-public import Iris.ProofMode.InstancesUpdates
+public import Iris.ProofMode
 
 @[expose] public section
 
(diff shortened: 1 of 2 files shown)


Iris/Iris/BI/Lib/BUpdPlain.lean
```diff
@@ -87,26 +92,24 @@ theorem BUpd_BUpdPlain [BIUpdate PROP] [BIBUpdateSbi PROP] [BIAffine PROP] {P :
...
   iapply Hx $$ HP
 
--- FIXME: @[rocq_alias own_updateP] duplicate alias
-/-- We get the usual rule for frame preserving updates if we have an [own]
-  connective satisfying the following rule w.r.t. interaction with plainly. -/
+-- FIXME: no `@[rocq_alias own_updateP]`: the short name is already claimed by
+-- `base_logic/lib/own.v`'s `own_updateP`, aliased to `Iris.iOwn_updateP`.
+/-- We get the usual rule for frame preserving updates if we have an `own`
+connective satisfying the following rule w.r.t. interaction with plainly. -/
 theorem own_updateP [UCMRA M] {own : M → PROP} {x : M} {Φ : M → Prop}
```
Kaptch, line 99: What is the purpose of this lemma? It's not part of bupd_alt, and not used anywhere.

markusdemedeiros: Is it not the one on line 70? https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/base_logic/bupd_alt.v#L70

Kaptch: Oh, it is! Then I'm a bit confused, does comment mean that the name is exactly the same? And we have `ownM_updateP` in a different namespace? Is there a different order of dependencies?

markusdemedeiros: The Rocq names upstream are the same, and this cases issues for `rocq_alias` (to the best of my understanding of the `rocq_alias` script). We will have to revisit this missing alias eventually but I've seen it a few times and want to get them all at once.

Kaptch: Aaaa, it's the Rocq version has the same name in both cases, then all is good.

Kaptch: I copied the name and clicked on a wrong entry, when looking at the Rocq version, sorry.

PR #639, opened 2026-08-13
fix: merging and splitting fractions with `IsOp`

### LINKED ISSUE #637: `IsOp` diverges from Rocq
The behaviour of `IsOp` diverges from Rocq. See the exampled below.

```
import Iris.Algebra.Frac
import Iris.ProofMode.SynthInstance

open Iris Iris.CMRA Iris.ProofMode

def Qp.quarter : Qp := ⟨1 / 4, by grind⟩

def Qp.threeQuarters : Qp := ⟨3 / 4, by grind⟩

/- Issue: `isOpSplit_op` is not considered -/
#ipm_synth (IsOp (α := Qp) .split (Qp.threeQuarters + Qp.quarter) _ _)

-- solution: IsOp IsOp.Direction.split (Qp.threeQuarters + Qp.quarter) (Qp.threeQuarters + Qp.quarter).half
--   (Qp.threeQuarters + Qp.quarter).half

-- should be 
-- solution: IsOp IsOp.Direction.split (Qp.quarter • Qp.threeQuarters) Qp.quarter Qp.threeQuarters

```

cc @alvinylt

alvinylt:
## Description

Fixes #637 and an additional bug with `IsOp`.

The sum `q1 + q2` should be split as `q1` and `q2` instead of `q1.half` and `q2.half`.

Currently, `isOpSplit_op` in `IsOp.lean` corresponds to `is_op_lr_op` in Iris-Rocq and has higher priority (`default + 100`) than `isOpFrac_half`. This instance is expressed using the CMRA operation `•`:

```lean4
/-- info:
  solution: IsOp IsOp.Direction.split (Qp.threeQuarters • Qp.quarter) Qp.threeQuarters Qp.quarter,
  new goals: []
-/
#guard_msgs (whitespace := lax) in
#ipm_synth (IsOp .split (Qp.threeQuarters • Qp.quarter) _ _)
```

The problem is that Lean does not unfold the sum (`+`) into the operation (`•`) automatically during type class synthesis, so `isOpSplit_op` is not used.

The solution is to have an additional instance for this with a *higher* priority than `isOpFrac_half`. On the contrary, two `q.half` are merged into `q` instead of `q.half + q.half`. The instance for merging has a *lower* priority than `isOpFrac_half`.

Furthermore, `ipm_backtrack` is necessary for `isOp_pair_core_id_r` or otherwise `isOp_pair_core_id_l` is never reached.

## Checklist

### DIFF
--- a/Iris/IrisTest/Instances.lean
+++ b/Iris/IrisTest/Instances.lean
@@ -408,7 +408,7 @@ variable (q q1 q2 : Qp)
   new goals: []
 -/
 #guard_msgs (whitespace := lax) in
-#ipm_synth (IsOp .split (q1 + q2 : Qp) _ _)
+#ipm_synth IsOp .split (q1 + q2 : Qp) _ _
 
 /- Splitting a CMRA operation: `isOpFrac_split` is used instead of `isOpFrac_half`. -/
 /-- info:
@@ -416,7 +416,7 @@ variable (q q1 q2 : Qp)
   new goals: []
 -/
 #guard_msgs (whitespace := lax) in
-#ipm_synth (IsOp .split (q1 • q2) _ _)
+#ipm_synth IsOp .split (q1 • q2) _ _
 
 /- Splitting a `Qp` value, where `isOpFrac_split` is not applicable: use `isOpFrac_half`. -/
 /-- info:
@@ -424,7 +424,7 @@ variable (q q1 q2 : Qp)
   new goals: []
 -/
 #guard_msgs (whitespace := lax) in
-#ipm_synth (IsOp .split q _ _)
+#ipm_synth IsOp .split q _ _
 
 /- Merging two `Qp` values: `isOpFrac_half` is not applicable, use `isOpFrac_merge`. -/
 /-- info:
@@ -432,7 +432,7 @@ variable (q q1 q2 : Qp)
   new goals: []
 -/
 #guard_msgs (whitespace := lax) in
-#ipm_synth (IsOp .merge _ q1 q2)
+#ipm_synth IsOp .merge _ q1 q2
 
 /- Merging two `Qp` values: `isOpFrac_half` is applicable and preferred for eliminating `.half`. -/
 /-- info:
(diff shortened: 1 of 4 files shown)


Iris/IrisTest/Instances.lean
```diff
@@ -395,3 +396,67 @@ variable (m n p q : Nat)
 #ipm_synth (NatCancel (1 + m + 2) 3 _ _ _)
 
 end NatCancel
+
+section IsOp
+open Iris Iris.CMRA Iris.ProofMode
+
+private def Qp.quarter : Qp := ⟨1 / 4, by grind⟩
+private def Qp.threeQuarters : Qp := ⟨3 / 4, by grind⟩
```
MackieLoeffel: Does it make sense to move these definitions to `Qp`?

MackieLoeffel (approved):
Looks good to me, except for one minor comment. Thanks for fixing this!

Iris/IrisTest/Instances.lean
alvinylt: Do you mean placing it alongside `Qq.half`? If so, we might want to add additional instances so that `q.quarter` and `q.threeQuarters` merge into `q` (likewise for merging `q.threeQuarters` and `q.quarter`). Not sure if they are practically useful/necessary. The tests can also be expressed in terms of generic `Qp` values.

Commit: https://github.com/leanprover-community/iris-lean/pull/639/commits/abe8968cb10788ab8ef2b85a42c48b71bc35a8d5

MackieLoeffel: This PR https://github.com/leanprover-community/iris-lean/pull/608 actually already introduced quarter and three quarters, so maybe we can use them here?

alvinylt: Sure, two more instances have been added along with tests.

Commits: https://github.com/leanprover-community/iris-lean/pull/639/commits/1c62d63e2e8f9dad80f7fdca0c871e87d599d6b8, https://github.com/leanprover-community/iris-lean/pull/639/commits/7fa947102c0c8daa4ade03362340f9402656e3b5

MackieLoeffel:
Great, thank you! Merging.

PR #645, opened 2026-08-14
Cleanup of OFE in preparation of the generalization of stepindices

MackieLoeffel:
## Description
This PR contains the changes from #576 without the actual generalization of OFE by the stepindex. This way, we can delay making a decision of the strategy of the step index generalization.


markusdemedeiros (approved):
Looks good!

Issue #251, opened 2026-04-07
Port proofmode/class_instances.v


lzy0505:
## Summary

Port the remaining items from [`proofmode/class_instances.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances.v).

> **Note:** Corresponds to `Instances.lean` in the Lean port.

## Subtasks

- [ ] Instances for telescopes
- [ ] Instances for big ops
- [x] MaybeCombineSepAs instances
- [x] CombineSepGives instances
- [x] AddModal instances
- [x] ElimInv instances

## Already ported

- [x] Instances for basic connectives
- [x] ElimModal instances

## Dependencies

**Rocq dependencies:**

- #211
- #219
- #212
- #239
- #225

alvinylt:
All subtasks except those involving `BigOp` and telescopes are done by #650.

Issue #224, opened 2026-04-07
Port proofmode/class_instances_frame.v


lzy0505:
## Summary

Port [`proofmode/class_instances_frame.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/class_instances_frame.v).

## Dependencies

**Rocq dependencies:**

- #211
- #219
- #212

PR #659, opened 2026-08-18
feat: proof mode instances for telescopes

alvinylt:
## Description

Since the telescopes have been implemented (#598), their proof mode instances can now also be ported.

Addresses #251 and #224 in part.

## Checklist


Iris/IrisTest/Telescopes.lean
```diff
@@ -0,0 +1,126 @@
...
+  iintro Hwand ⟨%x, HΦ⟩
+  iexists x
+  iapply Hwand $$ HΦ
+
+/- Tests `frame_tforall` and `frame_texist`. -/
+/--
+  □HR : R
+  ∗HΦ : ∀.. x, Φ x
+  ∗HΨ : ∃.. y, Ψ y
+  ⊢ tforall Φ ∗ texist Ψ
```
MackieLoeffel: Something seems to be broken with the notation printing.

Iris/Iris/Std/Telescopes.lean
```diff
@@ -67,6 +67,38 @@ def fold {B : Type v} (step : (A : Type u) → (A → B) → B) : {TT : Tele.{u}
   | .nil, f => ULift.down f
   | .cons _, f => step _ fun x => fold step (f x)
 
+/-- Telescopic universal quantification at `Prop`. -/
+def tforall : {TT : Tele.{u}} → (TT.Arg → Prop) → Prop
```
MackieLoeffel, line 74: Can you please add notation for `tforall` and `texist` like in Rocq? (see https://gitlab.mpi-sws.org/iris/stdpp/-/blob/master/stdpp/telescopes.v#L224 )

MackieLoeffel (commented):
Thanks! Except for some comments about notations, this looks good to me.

Iris/Iris/Std/Telescopes.lean
alvinylt: Notations and delaborators added in commit https://github.com/leanprover-community/iris-lean/pull/659/commits/a91f8dbff0b970a49b51524bf77dc435d3ae4ee2.

Iris/IrisTest/Telescopes.lean
alvinylt: The `delab_rule` declarations only accounted for `tforall (fun x => Ψ x)` but not `tforall Ψ`. It would be problematic to simply extend the `delab_rule` definitions due to binder shadowing. Instead, it is now implemented using `app_delab` to choose a fresh binder name when necessary.

Commit: https://github.com/leanprover-community/iris-lean/pull/659/commits/3fd29a4e0e250ec1b921e1b92d40be56f19f0435

alvinylt: As a side note, most of the BI propositions are delaborated with the `iprop(...)` wrapper.

Example:

```lean4
delab_rule BIBase.and
  | `($_ $P $Q) => do ``(iprop($(← unpackIprop P) ∧ $(← unpackIprop Q)))
```

Is this intentional and necessary? I would think `iprop(...)` is useful for elaboration but not so much for pretty printing.
While the wrapper is stripped within the Iris Proof Mode either way, it does show up in some tests:

```lean4
/-- info:
  solution: IntoWand false false iprop(∀.. x, Φ x -∗ Ψ x)
    (WandMode.matching WandMode.Side.result) (Φ a) (Ψ a),
  new goals: []
-/
#guard_msgs (whitespace := lax) in
#ipm_synth @IntoWand PROP _ false false (tforall fun x => iprop(Φ x -∗ Ψ x))
  (.matching .result) _ (Ψ a)
```

MackieLoeffel: Sounds great :+1: 

> Is this intentional and necessary? I would think iprop(...) is useful for elaboration but not so much for pretty printing.

It is also important for pretty printing since one usually wants the result of the pretty printing to be parsable again. So it is important to have `IntoWand false false iprop(∀.. x, Φ x -∗ Ψ x)` instead of `IntoWand false false (∀.. x, Φ x -∗ Ψ x)` since the latter would not parse correctly and look like the pure tforall instead of `BI.tforall`.

Iris/Iris/BI/Telescopes.lean
```diff
@@ -39,14 +39,51 @@ macro_rules
...
+  else
+    -- Print `tforall Ψ` as `∀.. x, Ψ x`
+    let Ψ := e.appArg!
+    let dom := (← Meta.inferType Ψ).bindingDomain!
+    -- Rename the binder if `Ψ` already refers a binder of the same name to avoid capture
+    let n ← getUnusedName `x Ψ
+    Meta.withLocalDeclD n dom fun _ => do
+      let f ← SubExpr.withAppArg delab
+      let x := mkIdent n
+      `(iprop(∀.. $x:ident, $f $x))
```
MackieLoeffel: Hm, this is more complex that I had hoped it would be and more complex than the previous version. It also seems like `BI.forall` does not handle this kind of delaboration. Not sure if it is worth it to have the delaboration for `tforall Ψ` . Do you have thoughts about this?

Iris/IrisTest/Telescopes.lean
alvinylt: Sure, let's leave the `delab_rule` definitions for the BI propositions unchanged then.

Iris/Iris/BI/Telescopes.lean
alvinylt: The complication comes mostly from binder scoping and the possibility of duplicate names. Since the delaborator is only relevant to syntax printing, I reckon it is unlikely that this will complicate any downstream usage of `BI.tforall`/`BI.texist`. The notation looks nice though; shall we factor the common delaboration steps out and share it amongst `BIBase.forall`, `BIBase.exist`, `BI.tforall` and `BI.texist`?

alvinylt: A generic delaborator function is created and shared by the four quantifiers. A side benefit of using this delaborator is the removal of some unnecessary parentheses. The delaborator also obeys the configuration `pp.explicit` by leaving the notations untouched when the option is set as `true`.

Commit: https://github.com/leanprover-community/iris-lean/pull/659/commits/80bf5e773a974b24abce1177e51b2a6837760ce3

alvinylt: The delaborator function is now also used by `Std.Tele.tforall` and `Std.Tele.texist`, so it is shared by six quantifiers in total. We no longer delaborate non-lambda expressions (e.g. `tforall Ψ`) to stay consistent with the built-in first-order quantifiers.

Commits: https://github.com/leanprover-community/iris-lean/pull/659/commits/f7657644153a1f176a7383b8f1a9ef0c82c889d4, https://github.com/leanprover-community/iris-lean/pull/659/commits/b8b3a862bd4f123fca6190fb1fe87858a21becd4, https://github.com/leanprover-community/iris-lean/pull/659/commits/8545bcc75a6cbead299b1e990a5c86c9ae895b16

Iris/Iris/Std/Telescopes.lean
```diff
@@ -90,52 +88,19 @@ macro "∀.." xs:explicitBinders ", " P:term : term => do
 macro "∃.." xs:explicitBinders ", " P:term : term => do
   return ⟨← expandExplicitBinders ``texist xs P⟩
 
-/-- A delaborator for the telescopic universal quantifier. -/
+/-- A delaborator for the telescopic universal quantifier for `Prop`. -/
```
MackieLoeffel: Why "for Prop"?

MackieLoeffel (approved):
Except for one minor question, this looks good to me.

Iris/Iris/Std/Telescopes.lean
alvinylt: I was referring to `Iris.Std.Tele.tforall : {TT : Tele.{u}} → (TT.Arg → Prop) → Prop` that this delaborator handles, as opposed to `Iris.BI.tforall [BI PROP] {TT : Tele} (Ψ : TT.Arg → PROP) : PROP` for Iris-level entailments. The distinction should nonetheless be clear even without the comments, so I have reverted the change.

Commit: https://github.com/leanprover-community/iris-lean/pull/659/commits/a29a815e1c6bfefd700bea958cc3b7079cbd482e

Issue #603, opened 2026-08-10
`wp_rec` splits the goal via `wp_bind` where Rocq keeps it flat


kdvkrs:
Surfaced while working on `wp_apply`. Related to #480.

`wp_rec` reduces the redex under a `wp_bind`, so the surrounding evaluation context ends up
in the postcondition instead of staying in the WP expression:

```lean
-- (l ↦ some v) ⊢ WP hl((#(); !v(#l)) + #1) @ s ; E {{ _r, True }}
  wp_seq
-- Lean: ⊢ WP hl(!#l) {{ v, WP hl((v(&v) + #1)) {{ _r, True }} }}
-- Rocq:   WP ! #l + #1 {{ _, True }}
```

Visible whenever the evaluation context is non-empty and the body does not reduce to a value. Downstream
goals then differ from Rocq's.

**Cause.** `Iris/Iris/HeapLang/ProofMode.lean:378`

```lean
macro "wp_rec" : tactic =>
  `(tactic | (wp_bind _ _; iapply $(mkIdent `wp_rec):ident; rfl; imodintro; wp_finish))
```

Rocq's `wp_rec` does two things: (a) it takes the beta step *in context*, leaving one flat WP,
and (b) it unfolds the head constant to a rec-value, which `wp_pure`/`wp_pures` deliberately
never do (so `wp_pures` terminates on recursive functions). (Rocq's `AsRecV_recv` assert is
Ltac plumbing for (b).) The current implementation gets (b) right, since `iapply`/`rfl` work up to
defeq, but loses (a): nothing puts `K` back. Routing `wp_rec` through
`wp_pure (App _ _)` like Rocq would invert the trade: `tac_wp_pure` restores `K`, but instance
synthesis will not unfold a `def` to a `RecV`. 

**Affects** `wp_rec`, `wp_lam`, `wp_let`, `wp_seq`, `wp_match`, and the `try wp_seq`
fast-forward in `wp_store`/`wp_xchg` from #600.

Happy to fix this myself.

kdvkrs:
I talked with @MackieLoeffel about this and he suggested that `wp_seq` and consorts should
not go through `wp_rec` (because no unfolding is needed) and
use `wp_pure (_ _)` instead. This fixes the example above.

However, it doesn't fully address the issue: `wp_rec` still diverges from its Rocq counterpart when the head is a
defined function since it goes through `wp_bind`. Consider

```lean
def readIt : Val := hl_val(λ l, ! l)

[...]

Issue #480, opened 2026-06-19
Port HeapLang tactics


suhr:
Port remaining HeapLang tactics from Rocq Iris.

Tactics to take one or more pure program steps:

- [x] wp_pure
- [x] wp_pures
- [x] wp_rec
- [x] wp_lam
- [x] wp_let
- [x] wp_seq
- [x] wp_proj
- [x] wp_if_true
- [x] wp_if_false
- [x] wp_if
- [x] wp_unop
- [x] wp_binop
- [x] wp_op
- [x] wp_case
- [x] wp_match
- [x] wp_inj
- [x] wp_pair
- [x] wp_closure

Tactics for the heap:

- [x] wp_alloc
- [x] wp_load
- [x] wp_store
- [x] wp_cmpxchg_suc
- [x] wp_cmpxchg_fail
- [x] wp_cmpxchg
- [x] wp_faa

Further tactics:

- [x] wp_bind
- [x] wp_apply
- [x] wp_smart_apply

kdvkrs:
Claiming the heap tactics.

fresheed:
@kdvkrs I'm going to work on porting arrays part of `derived_laws.v` if you haven't started yet?
#526

PR #661, opened 2026-08-18
feat: `wp_apply` and `wp_smart_apply`

kdvkrs:
## Description

This PR ports Rocq's `wp_apply` and `wp_smart_apply`. `wp_apply lem` poses `lem` whose conclusion must be a `WP` and applies it at the first evaluation context decomposition that fits and pushes the ectx into the postcondition. `wp_smart_apply lem` retries after single pure steps until the lemma applies. Both accept specialisation patterns (via `$$`) and a `with` clause for subsequent `iintros`. 

This addresses the remaining parts of #480.

The PR consists of four commits that fall into distinct files:

| commit | file |
|---|---|
| [feat: outermost-first evaluation context search](https://github.com/leanprover-community/iris-lean/commit/b6d906ae036ed537294cab706bc80489130d3882) | `HeapLang/Tactic.lean` |
| [refactor: make `iApplyCore` public](https://github.com/leanprover-community/iris-lean/commit/71988a9c7a3ba6895f92b86632861643b5a66160) | `ProofMode/Tactics/Apply.lean` |
| [feat: implement `wp_apply` and `wp_smart_apply`](https://github.com/leanprover-community/iris-lean/commit/7aa3b9397c32a839dbd6c7d6e60e832ffb64eae1) | `HeapLang/ProofMode.lean`, tests |
| [refactor: use `wp_apply` in HeapLang examples](https://github.com/leanprover-community/iris-lean/commit/41ac4250f9f1ed31ca1349456caa338d4036eaef) | ten `HeapLang/Lib/` files |

The first commit adds an outermost-first sibling to `findECtx`, which — contrary to Rocq's `reshape_expr` — is innermost-first (see discussion below). The second commit makes `iApplyCore` public, the third commit implements `wp_apply` and `wp_smart_apply`. Finally, the fourth commit refactors the existing HeapLang examples to use `wp_apply` (31 call sites, -55 lines of boilerplate in the HeapLang example files).

I chose to give `wp_smart_apply` a fuel (`smartApplyFuel`) so it doesn't diverge. I don't think this can occur in practice besides inline syntactically recursive self-application (like `WP hl((rec f x := f x) #1) {{ Φ }}`), hence I also picked a pretty high value (~60s on my machine with heartbeats disabled).

### Evaluation context search order

As outlined, `findECtx` is *innermost-first* contrary to its Rocq counterpart (`reshape_expr`). For most tactics, this does not matter since there's usually only a single matching site for call-by-value redexes. Where it *does* matter is a search that commits, like the current implementation of `wp_rec` (as outlined in #603) that runs `wp_bind (_ _)` which matches *any application* and commits before knowing whether the subsequent `iapply` succeeds. Incidentally, this design plays well with the current innermost-first `findECtx` since sites like `f x y = App (App f x) y` would be decomposed inside-out, but would break on a global change to outermost-first. With the `wp_rec` reimplementation from #657 both decomposition orders should work in all currently existing tactics.

`wp_apply` however makes a point for outermost-first decomposition order: Considering an expression-polymorphic lemma like `wp_wand` where `iapply` succeeds in every decomposition, I argue that the user would expect it to apply to the outermost expression instead of the innermost one. For example, `wp_apply wp_wand` on `WP (!#l + #1) {{ Φ }}` should, in my opinion, apply `wp_wand` on the whole expression and not the innermost `#l` which is not even a redex. There is however a point to be made about pinning the premises of `wp_apply`'d lemmas such that the expression is fixed before the search runs (which is what happens most of the time in the iris developments I scoured), or using `iapply` directly when wanting to target the whole expression. 

The use of `findECtxOutermost` in `wp_apply` only is certainly debatable and I'd appreciate the maintainer's input on this question.

## Checklist

## Generative AI Guidelines

### DIFF
--- a/Iris/Iris/HeapLang/ProofMode.lean
+++ b/Iris/Iris/HeapLang/ProofMode.lean
@@ -469,57 +469,81 @@ meta def iWpApplyCore {u} {GF : Q(BundledGFunctors.{0, 0, 0})} {hlc : Q(HasLC)}
     {prop : Q(Type u)} {bi : Q(BI $prop)} {ehyps : Q($prop)}
     (hyps : Hyps bi ehyps) (ι : Q(IrisGS_gen $hlc Exp $GF)) (s : Q(Stuckness)) (E : Q(CoPset))
     (e : Q(Exp)) (Φ : Q(Val → $prop)) (pmt : PMTerm)
-    (premisesOut : IO.Ref (Array MVarId × Array MVarId))
     (_hu : QuotedLevelDefEq u 0 := ⟨⟩) (_hprop : $prop =Q IProp $GF := ⟨⟩)
     (_hbi : $bi =Q UPred.instBIUPred := ⟨⟩)
     (κ : Q(Wp $prop Exp Val Stuckness) := q(wp.def)) (_hwp : $κ =Q wp.def := ⟨⟩) :
     ProofModeM Q($ehyps ⊢ Wp.wp $s $E $e $Φ) := do
-  let beforePose := (← get).goals.size
   let ⟨eΔ, hyps', p, A, posePf⟩ ← iHave hyps q(Wp.wp $s $E $e $Φ) pmt true
-  let afterPose := (← get).goals.size
   let some {result := pf, ..} ←
     findECtx (α := Q($eΔ ∗ □?$p $A ⊢ Wp.wp $s $E $e $Φ)) e fun K e' => do
       trace[wp_apply] m!"trying to apply {A} to {e'}"
       iWpBindCore _ ι s E e Φ K e' (iApply hyps' p A ·)
   | throwIPMError "cannot apply {A}"
-  premisesOut.set ((← get).goals.extract afterPose, (← get).goals.extract beforePose afterPose)
   return q($posePf $pf)
 
-/-- Strip a leading `▷` and simplify WP expressions, in the application's goals only -/
-meta def wpApplyPostPass (premises : Array MVarId) : TacticM (Array MVarId) := do
-  let post ← `(tactic| (try inext; try wp_expr_simp))
-  let mut out : Array MVarId := #[]
-  let mut produced : Array MVarId := #[]
-  for g in (← getGoals) do
-    if premises.contains g && !(← g.isAssigned) then
-      let new := (← Tactic.run g (evalTactic post)).toArray
-      out := out ++ new
-      produced := produced ++ new
+-- /-- Strip a leading `▷` and simplify WP expressions, in the application's goals only -/
+-- meta def wpApplyPostPass (premises : Array MVarId) : TacticM (Array MVarId) := do
+--   let post ← `(tactic| (try inext; try wp_expr_simp))
+--   let mut out : Array MVarId := #[]
+--   let mut produced : Array MVarId := #[]
+--   for g in (← getGoals) do
+--     if premises.contains g && !(← g.isAssigned) then
+--       let new := (← Tactic.run g (evalTactic post)).toArray
+--       out := out ++ new
+--       produced := produced ++ new
+--     else
+--       out := out.push g
+--   setGoals out.toList
+--   return produced
+
+-- /-- Apply and post-pass, return the application's goals. Dependee mvars (a lemma's `?Φ`)
+-- are registered before the snapshot, so they stay out of `premises` despite sorting last. -/
+-- meta def runWpApply (tacName : Name) (pmt : PMTerm) :
+--     TacticM (Array MVarId × Array MVarId) := do
+--   let premises ← IO.mkRef ((#[], #[]) : Array MVarId × Array MVarId)
+--   ProofModeM.runTacticWp tacName fun mvar {hyps, ι, s, E, e, Φ, ..} => do
+--     mvar.assign (← iWpApplyCore hyps ι s E e Φ pmt premises)
+--   let (application, specialisation) ← premises.get
+--   return (← wpApplyPostPass application, specialisation)
+
+-- /-- Needed for `wp_apply ... with`. Introduce into the last goal the application produced,
+-- or a `$$` goal if it produced none. -/
+-- meta def wpApplyIntro (tacName : Name) (produced specialisation : Array MVarId)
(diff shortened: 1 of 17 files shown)


Iris/IrisTest/HeapLang/Apply.lean
```diff
@@ -0,0 +1,349 @@
...
+-/
+#guard_msgs (whitespace := lax) in
+example {Q R : IProp GF} {e : Exp} {Φ : Val → IProp GF} : ⊢@{IProp GF}
+    ((Q -∗ R) -∗ WP hl(&e) {{ Φ }}) -∗ (Q -∗ R) -∗ WP hl(&e) {{ Φ }} := by
+  iintro H HQR
+  wp_apply H $$ [HQR] with HQ
+
+-- the continuation handed to a `$$` pattern
+/--
+error: unsolved goals
```
alvinylt: A minor remark: shall we use `trace_state` in place of `error: unsolved goals` in these tests?

Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -447,6 +461,154 @@ macro "wp_pair" : tactic => `(tactic | wp_pure ((_, _)))
...
+/--
+`wp_smart_apply lem` is like `wp_apply lem`, but when the lemma does not apply,
+it takes single pure steps (`wp_pure`) and retries, until the lemma applies or
+no pure step is possible. `$$` and `with` behave as for `wp_apply`. The retry is
+bounded under `smartApplyFuel` so the tactic does not diverge.
+-/
+syntax (name := wpSmartApply) "wp_smart_apply " colGt pmTerm
+  (" with" (colGt ppSpace introPat)+)? : tactic
+
+elab_rules : tactic
```
MackieLoeffel: The implementation of `wp_smart_apply` is broken after my refactoring. @alvinylt Would it be possible for you to look into making `wp_smart_apply` work again?

alvinylt: Sure, I'll have a look into it.

MackieLoeffel: Thanks!

There are two options for implementing `wp_smart_apply`: The first is to do it at the tactic level, by composing the `wp_apply_raw` tactic with other tactics. This is also what the current PR tries to do. It is probably the simpler version, but it does not allow very finegrained error messages and might do more backtracking than necessary. (In particular, the implementation will likely use `ihave` multiple times on the proof mode term.)

The other option is to implement `wp_smart_apply` at the ProofModeM level. Probably by adding a loop to `iWpApplyCore` with a parameter describing how it should be retried. This is in spirit more similar to the Rocq implementation and avoids some of the problems of the tactic based version. E.g. one can make sure that a wrong specialization pattern does not trigger a wp_pure, but directly aborts the tactic. So I think this option would be preferred, but it likely requires some tweaks to the current ProofModeM functions in this file.

I leave it up to you, which version you think is better / is easier to implement.

alvinylt: The tactic `wp_smart_apply` was broken because `runWpSmartApply` used `wp_apply_raw` directly rather than `wp_apply_raw $pmt <;> (try inext) <;> (try wp_expr_simp)`.

The first solution is to factor out the extra steps, i.e., `(try inext) <;> (try wp_expr_simp)` (commit https://github.com/kdvkrs/iris-lean/commit/423ceb6ef450f683e56a088ce14799cd7214c975) and use it for not only `wp_apply` but also `wp_smart_apply`.

For the second (and most recent) approach, `iWpPureCore` is extended with a continuation function `k`, which allows `iWpApplyCore` to make recursive calls when `wp_smart_apply` is used. Both approaches manage to pass all the tests. The second solution feels more elegant, as the retries are contained within one `ProofMode.runTactic` call. I don't have write permissions for this PR, so the fixes are pushed to [a separate branch](https://github.com/kdvkrs/iris-lean/compare/feat/wp-apply...ISTA-PLV:iris-lean:feat/wp-apply).

Iris/IrisTest/HeapLang/Apply.lean
alvinylt: Updated in commit https://github.com/kdvkrs/iris-lean/commit/48f6ca5b5772b4eb956f1d4b14cb24ca2104de99.

Iris/Iris/HeapLang/ProofMode.lean
```diff
@@ -447,6 +471,134 @@ macro "wp_pair" : tactic => `(tactic | wp_pure ((_, _)))
...
+    (κ : Q(Wp $prop Exp Val Stuckness) := q(wp.def)) (_hwp : $κ =Q wp.def := ⟨⟩) :
+    ProofModeM Q($ehyps ⊢ Wp.wp $s $E $e $Φ) := do
+  let mut st : @WpApplyState u GF hlc prop bi ehyps s E e Φ κ :=
+  { hypsC := hyps, eC := e,
+    prefixPf := q(fun (pf : $ehyps ⊢ @Wp.wp $prop Exp Val Stuckness $κ $s $E $e $Φ) => pf) }
+  let mut firstFailed : Option MessageData := none
+  repeat
+    let ⟨hypsC, eC, prefixPf⟩ := st
+    let saved ← ProofModeM.saveState
+    let ⟨ehypsP, hypsP, p, A, posePf⟩ ← iHave hypsC q(Wp.wp $s $E $eC $Φ) pmt true
```
MackieLoeffel: You moved the `iHave` back into the loop. Would it be possible to keep it outside of the loop such that it only happens once at the beginning instead of at every iteration? This is also how the Rocq version works.

alvinylt: Refactored to have `iHave` before entering the loop. This means the mutable state begins with the proof state that includes the result of `iHave`.

Commit: https://github.com/leanprover-community/iris-lean/commit/ea18ee62087f1cfc5ad7f3b3705948746f131b13

alvinylt: The code `save.restore ...` is also removed, as there is no longer any need to handle failed poses within the loop.

PR #663, opened 2026-08-18
feat: port bi/Algebra and sbi_unfold

markusdemedeiros:
## Description
Finishes `Algebra.v`. 

There are some lightweight tactics for dealing with `sbi` simplifications. `sbi_unfold.v` does this in Rocq, I'm not sure I understand what they're doing enough to determine if we should replicate them or ignore them. The tactics here seem to work alright. Polling you guys for your thoughts @Kaptch @MackieLoeffel. By your reading, is it doing anything that a `simp` set could not? 

## Checklist

## Generative AI Guidelines


Iris/Iris/BI/Algebra.lean
```diff
@@ -7,46 +7,61 @@ module
...
+      ← (Iris.siPure_pure (PROP := $p)).to_eq, ← (Iris.siPure_and (PROP := $p)).to_eq,
+      ← (Iris.siPure_or (PROP := $p)).to_eq, ← (Iris.siPure_exist (PROP := $p)).to_eq,
+      ← (Iris.siPure_forall (PROP := $p)).to_eq, ← (Iris.siPure_later (PROP := $p)).to_eq])
+
+/-- Unfold `SiProp.holds` of the connectives into their meta-level meaning. -/
+local macro "sbi_holds" : tactic =>
+  `(tactic| simp only [Iris.SiProp.pure_holds, Iris.SiProp.and_holds, Iris.SiProp.or_holds,
+      Iris.SiProp.sep_holds, Iris.SiProp.exists_holds, Iris.SiProp.forall_holds,
+      Iris.SiProp.later_holds_zero, Iris.SiProp.later_holds_succ,
+      Iris.SiProp.internalEq_holds, Iris.SiProp.cmraValid_holds, Iris.siPure_holds])
```
MackieLoeffel: Instead of the hard coded `simp only` list, I would use a custom simp attribute defined with `register_simp_attr`. This makes it easier to extend this list.
See `wp_expr_simp` for an example. 
(Not sure if this approach also works for `sbi_fold`. The `p` parameter might prevent one from using a custom simp attribute there.)

Iris/Iris/BI/SIProp.lean
```diff
@@ -246,6 +246,40 @@ instance instBI : BI SiProp where
...
 
+/-! ## Step-indexed characterisation of the connectives
+
+`BI`'s quantifiers range over *sets* of `SiProp`s, so `∃`/`∀` do not reduce to their
+meta-level counterparts by `rfl`; the remaining connectives do. -/
+
+theorem biEntails_of_iff {P Q : SiProp} (h : ∀ n, P.holds n ↔ Q.holds n) : P ⊣⊢ Q :=
+  ⟨fun n => (h n).mp, fun n => (h n).mpr⟩
+
+@[simp] theorem pure_holds {φ : Prop} {n} : (iprop(⌜φ⌝) : SiProp).holds n ↔ φ := .rfl
```
Kaptch, line 257: Would it be okay to add those to simpset as well instead of adding a simp attribute?

Iris/Iris/BI/Algebra.lean
markusdemedeiros: Yes, I think I might roll sbi_unfold into this PR and just do it properly. Finally an occasion to learn about simpprocs.

Iris/Iris/BI/SIProp.lean
markusdemedeiros: Probably a good idea. I'm going to redo the automation around `sbi` this morning and will re-request after.

MackieLoeffel (approved):
Looks good to me.

Kaptch (approved):

PR #677, opened 2026-08-22
feat: port `proofmode/monpred.v`

### LINKED ISSUE #281: Port proofmode/monpred.v
## Summary

Port [`proofmode/monpred.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/monpred.v).

## Dependencies

**Rocq dependencies:**

- `proofmode/proofmode` (aggregation)
- #210
- #266
- #212
- #225

alvinylt:
## Description

Ports [`proofmode/monpred.v`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/monpred.v), hence 100% progress for the `ProofMode` module.

Resolves #281.

Additionally, a few sets of changes to `BI/MonPred.lean`:
- [x] Ports the missing notations `<obj>` and `<subj>`
- [x] Ports the missing four theorems involving `BigSepMSet`
- [x] Code style improvements
  - Remove lots of redundant `iprop(...)` wrappers and parentheses
  - Remove redundant prefixes within the namespace `MonPred`
  - `calc`-style proofs for transitive BI entailments

### Checklist of Ported Items

<details>

<summary><b>Instances of existing type classes</b></summary>

- [x] `add_modal_at_bupd_goal`
- [x] `add_modal_at_fupd_goal`
- [x] `as_emp_valid_monPred_at`
- [x] `as_emp_valid_monPred_at_equiv`
- [x] `as_emp_valid_monPred_at_wand`
- [x] `elim_acc_at_None`
- [x] `elim_acc_at_Some`
- [x] `elim_inv_embed_with_close`
- [x] `elim_inv_embed_without_close`
- [x] `elim_modal_at`
- [x] `elim_modal_at_bupd_goal`
- [x] `elim_modal_at_bupd_hyp`
- [x] `elim_modal_at_fupd_goal`
- [x] `elim_modal_at_fupd_hyp`
- [x] `from_and_monPred_at`
- [x] `from_assumption_make_monPred_at_l`
- [x] `from_assumption_make_monPred_at_r`
- [x] `from_assumption_make_monPred_objectively`
- [x] `from_assumption_make_monPred_subjectively`
- [x] `from_exist_monPred_at`
- [x] `from_exist_monPred_at_ex`
- [x] `from_forall_monPred_at`
- [x] `from_forall_monPred_at_impl`
- [x] `from_forall_monPred_at_objectively`
- [x] `from_forall_monPred_at_plainly`
- [x] `from_forall_monPred_at_wand`
- [x] `from_later_monPred_at`
- [x] `from_modal_affinely_monPred_at`
- [x] `from_modal_id_monPred_at`
- [x] `from_modal_intuitionistically_monPred_at`
- [x] `from_modal_objectively`
- [x] `from_modal_persistently_monPred_at`
- [x] `from_modal_subjectively`
- [x] `from_or_monPred_at`
- [x] `from_pure_monPred_at`
- [x] `from_pure_monPred_in`
- [x] `from_sep_monPred_at`
- [x] `into_and_monPred_at`
- [x] `into_embed_objective`
- [x] `into_except_0_monPred_at_bwd`
- [x] `into_except_0_monPred_at_fwd`
- [x] `into_exist_monPred_at`
- [x] `into_exist_monPred_at_ex`
- [x] `into_forall_monPred_at`
- [x] `into_forall_monPred_at_index`
- [x] `into_forall_monPred_at_objectively`
- [x] `into_forall_monPred_at_plainly`
- [x] `into_internal_eq_monPred_at`
- [x] `into_or_monPred_at`
- [x] `into_persistent_monPred_at`
- [x] `into_pure_monPred_at`
- [x] `into_pure_monPred_in`
- [x] `into_sep_monPred_at`
- [x] `into_wand_impl'_monPred` (ignored)
- [x] `into_wand_monPred_at_known_unknown_ge`
- [x] `into_wand_monPred_at_known_unknown_le`
- [x] `into_wand_monPred_at_unknown_known`
- [x] `into_wand_monPred_at_unknown_unknown`
- [x] `into_wand_wand'_monPred` (ignored)
- [x] `is_except_0_monPred_at`
- [x] `maybe_into_later_monPred_at`

</details>

<details>

<summary><b><tt>FrameMonPredAt</tt> and its instances</b></summary>

- [x] `FrameMonPredAt`
- [x] `frame_monPred_at_absorbingly`
- [x] `frame_monPred_at_affinely`
- [x] `frame_monPred_at_and`
- [x] `frame_monPred_at_bupd`
- [x] `frame_monPred_at_embed`
- [x] `frame_monPred_at_enter`
- [x] `frame_monPred_at_exist`
- [x] `frame_monPred_at_forall`
- [x] `frame_monPred_at_fupd`
- [x] `frame_monPred_at_here`
- [x] `frame_monPred_at_impl`
- [x] `frame_monPred_at_intuitionistically`
- [x] `frame_monPred_at_later`
- [x] `frame_monPred_at_laterN`
- [x] `frame_monPred_at_objectively`
- [x] `frame_monPred_at_or`
- [x] `frame_monPred_at_persistently`
- [x] `frame_monPred_at_sep`
- [x] `frame_monPred_at_subjectively`
- [x] `frame_monPred_at_wand`

</details>

<details>

<summary><b><tt>MakeMonPredAt</tt> and its instances</b></summary>

- [x] `MakeMonPredAt`
- [x] `make_monPred_at_absorbingly`
- [x] `make_monPred_at_absorbingly_if`
- [x] `make_monPred_at_affinely`
- [x] `make_monPred_at_affinely_if`
- [x] `make_monPred_at_and`
- [x] `make_monPred_at_bupd`
- [x] `make_monPred_at_default`
- [x] `make_monPred_at_embed`
- [x] `make_monPred_at_emp`
- [x] `make_monPred_at_except_0`
- [x] `make_monPred_at_exists`
- [x] `make_monPred_at_forall`
- [x] `make_monPred_at_fupd`
- [x] `make_monPred_at_in`
- [x] `make_monPred_at_internal_eq`
- [x] `make_monPred_at_intuitionistically`
- [x] `make_monPred_at_intuitionistically_if`
- [x] `make_monPred_at_later`
- [x] `make_monPred_at_laterN`
- [x] `make_monPred_at_or`
- [x] `make_monPred_at_persistently`
- [x] `make_monPred_at_persistently_if`
- [x] `make_monPred_at_pure`
- [x] `make_monPred_at_sep`

</details>

<details>

<summary><b><tt>IsBiIndexRel</tt> and its instance</b></summary>

- [x] `IsBiIndexRel`
- [x] `is_bi_index_rel_refl`

</details>

<details>

<summary><b>Modality-related items</b></summary>

- [x] `modality_objectively`
- [x] `modality_objectively_mixin`

</details>

## Checklist

### DIFF
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -52,9 +52,10 @@ theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io ioP}
   inst.as_emp_valid.right rfl
 
 @[ipm_class, rocq_alias AsEmpValid0]
-class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut := d.toInOut)
+class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut)
     (PROP : semiOutParamIPM io (Type _))
-    (bi : semiOutParamIPM io (BI PROP)) (ioP := .out) (P : semiOutParamIPM ioP PROP) where
+    (bi : semiOutParamIPM io (BI PROP))
+    ioP (P : semiOutParamIPM ioP PROP) where
   as_emp_valid_0 : AsEmpValid d φ io PROP bi ioP P
 
 attribute [ipm_backtrack,instance] AsEmpValid0.as_emp_valid_0
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -227,7 +227,8 @@ class IsExcept0 {PROP} [BI PROP] (Q : PROP) where
 export IsExcept0 (is_except0)
 
 @[ipm_class, rocq_alias IntoExcept0]
-class IntoExcept0 {PROP} [BI PROP] (P : PROP) (Q : outParam PROP) where
+class IntoExcept0 {PROP} [BI PROP] (io : InOut)
+    (P : semiOutParamIPM io PROP) (Q : semiOutParamIPM io.negate PROP) where
   into_except0 : P ⊢ ◇ Q
 export IntoExcept0 (into_except0)
 
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -37,25 +37,25 @@ end
 @[ipm_class, rocq_alias AsEmpValid]
 class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) io
     (PROP : semiOutParamIPM io (Type _))
-    (bi : semiOutParamIPM d.toInOut (BI PROP))
-    (P : outParam PROP) where
+    (bi : semiOutParamIPM io (BI PROP))
+    ioP (P : semiOutParamIPM ioP PROP) where
   as_emp_valid : (d = .into → φ → ⊢ P) ∧ (d = .from → (⊢ P) → φ)
 
 @[rocq_alias as_emp_valid_1]
-theorem asEmpValid_1 {PROP} [bi : BI PROP] {φ : Prop} (P : PROP) {io}
-    (inst : AsEmpValid .into φ io PROP bi P) : φ → ⊢ P :=
+theorem asEmpValid_1 {PROP} [bi : BI PROP] {φ : Prop} (P : PROP) {io ioP}
+    (inst : AsEmpValid .into φ io PROP bi ioP P) : φ → ⊢ P :=
   inst.as_emp_valid.left rfl
 
 @[rocq_alias as_emp_valid_2]
-theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io}
-    (inst : AsEmpValid .from φ io PROP bi P) : (⊢ P) → φ :=
+theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io ioP}
+    (inst : AsEmpValid .from φ io PROP bi ioP P) : (⊢ P) → φ :=
   inst.as_emp_valid.right rfl
 
 @[ipm_class, rocq_alias AsEmpValid0]
 class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut := d.toInOut)
     (PROP : semiOutParamIPM io (Type _))
-    (bi : semiOutParamIPM d.toInOut (BI PROP)) (P : outParam PROP) where
-  as_emp_valid_0 : AsEmpValid d φ io PROP bi P
+    (bi : semiOutParamIPM io (BI PROP)) (ioP := .out) (P : semiOutParamIPM ioP PROP) where
+  as_emp_valid_0 : AsEmpValid d φ io PROP bi ioP P
 
 attribute [ipm_backtrack,instance] AsEmpValid0.as_emp_valid_0
 
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -38,17 +38,17 @@ end
 class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) io
     (PROP : semiOutParamIPM io (Type _))
     (bi : semiOutParamIPM io (BI PROP))
-    ioP (P : semiOutParamIPM ioP PROP) where
+    (P : outParam PROP) where
   as_emp_valid : (d = .into → φ → ⊢ P) ∧ (d = .from → (⊢ P) → φ)
 
 @[rocq_alias as_emp_valid_1]
-theorem asEmpValid_1 {PROP} [bi : BI PROP] {φ : Prop} (P : PROP) {io ioP}
-    (inst : AsEmpValid .into φ io PROP bi ioP P) : φ → ⊢ P :=
+theorem asEmpValid_1 {PROP} [bi : BI PROP] {φ : Prop} (P : PROP) {io}
+    (inst : AsEmpValid .into φ io PROP bi P) : φ → ⊢ P :=
   inst.as_emp_valid.left rfl
 
 @[rocq_alias as_emp_valid_2]
-theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io ioP}
-    (inst : AsEmpValid .from φ io PROP bi ioP P) : (⊢ P) → φ :=
+theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io}
+    (inst : AsEmpValid .from φ io PROP bi P) : (⊢ P) → φ :=
   inst.as_emp_valid.right rfl
 
 @[ipm_class, rocq_alias AsEmpValid0]
@@ -56,9 +56,13 @@ class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut)
     (PROP : semiOutParamIPM io (Type _))
     (bi : semiOutParamIPM io (BI PROP))
     ioP (P : semiOutParamIPM ioP PROP) where
-  as_emp_valid_0 : AsEmpValid d φ io PROP bi ioP P
+  as_emp_valid_0 : AsEmpValid d φ io PROP bi P
 
-attribute [ipm_backtrack,instance] AsEmpValid0.as_emp_valid_0
+@[ipm_backtrack]
+instance asEmpValid_of_asEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) io
+    (PROP : Type _) (bi : BI PROP) (P : PROP)
+    [inst : AsEmpValid0 d φ io PROP bi .out P] :
+    AsEmpValid d φ io PROP bi P := inst.as_emp_valid_0
 
 /- Depending on the use case, type classes with the prefix `From` or `Into` are used. Type classes
 with the prefix `From` are used to generate one or more propositions *from* which the original
--- a/Iris/Iris/ProofMode/Classes.lean
+++ b/Iris/Iris/ProofMode/Classes.lean
@@ -232,8 +232,7 @@ class IsExcept0 {PROP} [BI PROP] (Q : PROP) where
 export IsExcept0 (is_except0)
 
 @[ipm_class, rocq_alias IntoExcept0]
-class IntoExcept0 {PROP} [BI PROP] (io : InOut)
-    (P : semiOutParamIPM io PROP) (Q : semiOutParamIPM io.negate PROP) where
+class IntoExcept0 {PROP} [BI PROP] (P : PROP) (Q : outParam PROP) where
   into_except0 : P ⊢ ◇ Q
 export IntoExcept0 (into_except0)
 
--- a/Iris/Iris/ProofMode/InstancesEmbedding.lean
+++ b/Iris/Iris/ProofMode/InstancesEmbedding.lean
@@ -23,8 +23,8 @@ variable [bi1 : BI PROP1] [bi2 : BI PROP2] [BiEmbed PROP1 PROP2]
 set_option synthInstance.checkSynthOrder false in
 @[rocq_alias as_emp_valid_embed]
 instance (priority := low) asEmpValid_embed d φ io (P : PROP1)
-    [inst : AsEmpValid0 d φ io PROP1 bi1 ioP P] :
-    AsEmpValid d φ io PROP2 bi2 ioP (embed P) where
+    [inst : AsEmpValid0 d φ io PROP1 bi1 .out P] :
+    AsEmpValid d φ io PROP2 bi2 (embed P) where
   as_emp_valid := by
     constructor
     · exact fun hd hφ => (embed_emp_valid P).mpr <| inst.as_emp_valid_0.as_emp_valid.left hd hφ
--- a/Iris/Iris/ProofMode/InstancesEmbedding.lean
+++ b/Iris/Iris/ProofMode/InstancesEmbedding.lean
@@ -280,12 +280,11 @@ instance fromModal_later_embed [BiEmbedLater PROP1 PROP2] {α} φ io (sel : α) 
 
 /-! ### IntoExcept0 -/
 
-set_option synthInstance.checkSynthOrder false in
 /-- The metavariables `PROP1` and `bi1` are unified in the conclusion before the premise runs. -/
 @[rocq_alias into_except_0_embed]
-instance intoExcept0_embed [BiEmbedLater PROP1 PROP2] (P Q : PROP1) io
-    [inst : IntoExcept0 io P Q] :
-    IntoExcept0 (PROP := PROP2) io iprop(⎡P⎤) iprop(⎡Q⎤) where
+instance intoExcept0_embed [BiEmbedLater PROP1 PROP2] (P Q : PROP1)
+    [inst : IntoExcept0 P Q] :
+    IntoExcept0 (PROP := PROP2) iprop(⎡P⎤) iprop(⎡Q⎤) where
   into_except0 := (embed_mono inst.into_except0).trans (embed_except_0 Q).mp
 
 /-! ### IntoLater -/
--- a/Iris/Iris/ProofMode/InstancesEmbedding.lean
+++ b/Iris/Iris/ProofMode/InstancesEmbedding.lean
@@ -20,7 +20,6 @@ variable [bi1 : BI PROP1] [bi2 : BI PROP2] [BiEmbed PROP1 PROP2]
 
 /-! ### AsEmpValid -/
 
-set_option synthInstance.checkSynthOrder false in
 @[rocq_alias as_emp_valid_embed]
 instance (priority := low) asEmpValid_embed d φ io (P : PROP1)
     [inst : AsEmpValid0 d φ io PROP1 bi1 .out P] :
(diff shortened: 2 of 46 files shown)


Iris/Iris/ProofMode/MonPred.lean
```diff
@@ -0,0 +1,1131 @@
...
+
+set_option synthInstance.checkSynthOrder false in
+/--
+  `Q.monPred_at i` is an input argument of `IntoExcept0`.
+  Unifying the conclusion assigns the metavariables before the assumption is involved.
+-/
+@[rocq_alias into_except_0_monPred_at_bwd]
+instance intoExcept0_monPred_at_bwd (P Q : MonPred I PROP) (𝓟 : PROP) (i : I.car)
+    [h : IntoExcept0 .out P Q] [instMP : MakeMonPredAt .indexToProp i P 𝓟] :
+    IntoExcept0 .out 𝓟 (Q.monPred_at i) where
```
MackieLoeffel: Is the `.out` version actually used anywhere else? Or should this be `.in` instead of `.out` here? (In which the `io` is maybe not necessary, but we probably want some `ipm_backtrack` and some priorities.)

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -37,25 +37,25 @@ end
...
+theorem asEmpValid_2 {PROP} [bi : BI PROP] {P: PROP} (φ : Prop) {io ioP}
+    (inst : AsEmpValid .from φ io PROP bi ioP P) : (⊢ P) → φ :=
   inst.as_emp_valid.right rfl
 
 @[ipm_class, rocq_alias AsEmpValid0]
 class AsEmpValid0 (d : AsEmpValid.Direction) (φ : Prop) (io : InOut := d.toInOut)
     (PROP : semiOutParamIPM io (Type _))
-    (bi : semiOutParamIPM d.toInOut (BI PROP)) (P : outParam PROP) where
-  as_emp_valid_0 : AsEmpValid d φ io PROP bi P
+    (bi : semiOutParamIPM io (BI PROP)) (ioP := .out) (P : semiOutParamIPM ioP PROP) where
```
MackieLoeffel: I am not sure if this default parameter is a good idea since it might accidentally make instances less generic than desired. Is there a specific reason why you added it?

Iris/Iris/ProofMode/Classes.lean
```diff
@@ -37,25 +37,25 @@ end
 @[ipm_class, rocq_alias AsEmpValid]
 class AsEmpValid (d : AsEmpValid.Direction) (φ : Prop) io
     (PROP : semiOutParamIPM io (Type _))
-    (bi : semiOutParamIPM d.toInOut (BI PROP))
-    (P : outParam PROP) where
+    (bi : semiOutParamIPM io (BI PROP))
+    ioP (P : semiOutParamIPM ioP PROP) where
```
MackieLoeffel: Does `AsEmpValid` also need the new `io` parameters or is this only necessary for `AsEmpValid0`?

Iris/Iris/ProofMode/MonPred.lean
alvinylt: In Rocq, `IntoExcept0` is configured with the hint mode `! -` so it is more permissive than the existing definition of `IntoExcept0` in Lean. Lean complained about `𝓟` being a metavariable at an input argument position, that's why generalised `IntoExcept0` to work both ways.

Actually, I couldn't come up with a good example to test this instance beyond [the `#ipm_synth` line](https://github.com/ISTA-PLV/iris-lean/blob/7d44ac3027f76e8b6ac1a44d45ae6edf435bac3e/Iris/IrisTest/MonPred.lean#L312-L319), as the only real usage of `IntoExcept0` is [`elimModal_timeless`](https://github.com/ISTA-PLV/iris-lean/blob/7d44ac3027f76e8b6ac1a44d45ae6edf435bac3e/Iris/Iris/ProofMode/InstancesLater.lean#L381-L391) which only works at the forward direction. An alternative I considered was to ignore this instance, in which case there would be no need to change `IntoExcept0`.

Iris/Iris/ProofMode/Classes.lean
alvinylt: Agreed, default value for `ioP` removed.

Commit: https://github.com/leanprover-community/iris-lean/commit/4afd24a2b7b17ae9df18e4ac704b8d7d06066e97

Iris/Iris/ProofMode/Classes.lean
alvinylt: It also works with `P` fixed as an output parameter. The instance for embedding (`asEmpValid_embed`) is updated so that `ioP` for `AsEmpValid0` is fixed as `.out`:

```lean4
instance (priority := low) asEmpValid_embed d φ io (P : PROP1)
    [inst : AsEmpValid0 d φ io PROP1 bi1 .out P] :
    AsEmpValid d φ io PROP2 bi2 (embed P)
```

Commit: https://github.com/leanprover-community/iris-lean/commit/c6c960a884a1e61bea6b6e2be32724b7dc813628

Iris/Iris/ProofMode/MonPred.lean
MackieLoeffel: I think the point of the backward instance is to compute `P` from `𝓟` using `MakeMonPredAt` and then call `IntoExcept0` with `P` as the input. So I don't think `IntoExcept0` needs an `io` parameter, but maybe `MakeMonPredAt` does need to support `P` being an output.

alvinylt: If I understand correctly, `MakeMonPredAt` allows conversion between `𝓟` and `i`, while `P` is always an input? Considering the default instance [`makeMonPredAt_default`](https://github.com/ISTA-PLV/iris-lean/blob/50baa6571730df2e5cdb8704a17f977e68fa203e/Iris/Iris/ProofMode/MonPred.lean#L781-L784), would making `P` an output resulting in it matching any goal? Meanwhile, there are [two Hint Modes for `IntoExcept0`](https://gitlab.mpi-sws.org/iris/iris/-/blob/f0f9f3b67ec5a40ac51f4c8d8df23be89bc4913c/iris/proofmode/classes.v?blame=1#L414-415) which is a bit unusual, so I might understand it better with some concrete usage examples.

MackieLoeffel: Actually, what I said in my previous message about how the backward instance works might be wrong since I think the type class parameters are in the wrong order for this in the Rocq version. I don't understand how the backward instance is supposed to work in Rocq. Also I don't understand why IntoExcept0 has two hint modes in Rocq. It does not seem necessary in [the commit that introduces them](https://gitlab.mpi-sws.org/iris/iris/-/commit/7499050658800a43047a03e40e8f0a6b8f40593a).
Maybe we just ignore the backward instance for now?

alvinylt: Sure, let's ignore the instance and revert the change to `IntoExcept0` for the time being.

Commit: https://github.com/leanprover-community/iris-lean/pull/677/commits/edcd64afdbc34a33cdc7b7fcff7f543b65872cd7

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -333,16 +333,25 @@ instance fromModal_except0 [BI PROP] io (P : PROP) :
   from_modal _ := except0_intro
 
 /-! ### IntoExcept0 -/
+
+/- The three non-recursive instances require  -/
```
MackieLoeffel: ?

Iris/Iris/ProofMode/InstancesLater.lean
```diff
@@ -333,16 +333,25 @@ instance fromModal_except0 [BI PROP] io (P : PROP) :
...
+
+set_option synthInstance.checkSynthOrder false in
```
MackieLoeffel: Are these set_options still necessary?

Iris/Iris/ProofMode/InstancesEmbedding.lean
```diff
@@ -280,8 +280,10 @@ instance fromModal_later_embed [BiEmbedLater PROP1 PROP2] {α} φ io (sel : α)
 
 /-! ### IntoExcept0 -/
 
+/-- The metavariables `PROP1` and `bi1` are unified in the conclusion before the premise runs. -/
```
MackieLoeffel: I don't think this comment is necessary.

Iris/Iris/ProofMode/MonPred.lean
```diff
@@ -0,0 +1,1119 @@
...
+open Lean Meta
+
+/--
+  `MakeMonPredAt` is used by `AsEmpValid` to import lemmas into the proof mode,
+  with `makeMonPredAt_default` as the fallback option.
+  However, if any of `I`, `PROP` or `P` is a metavariable, `makeMonPredAt_default`
+  should not be used.
+-/
+@[ipm_tactic_instance MakeMonPredAt _ _ _ _]
+meta def makeMonPredAtModeGuard : SynthTactic := fun e => do
```
MackieLoeffel: Why is this necessary? Does the input checking of the ipm search not handle this?

Iris/Iris/ProofMode/MonPred.lean
```diff
@@ -0,0 +1,1119 @@
...
+  from_assumption := calc
+    _ ⊢ P.monPred_at j := intuitionisticallyIf_elim
+    _ ⊢ P.monPred_at i := P.monPred_mono instRel.is_bi_index_rel
+    _ ⊢ 𝓟              := instMP.make_monPred_at.mp
+
+@[ipm_backtrack, rocq_alias from_assumption_make_monPred_at_r]
+instance fromAssumption_make_monPred_at_r (p : Bool) d (i j : I.car)
+    (P : MonPred I PROP) (𝓟 : PROP)
+    [instMP : MakeMonPredAt d i P 𝓟] [instRel : IsBiIndexRel i j] :
+    FromAssumption p d.propIO 𝓟 (P.monPred_at j) where
```
MackieLoeffel: Does having `d.propIO` in the conclusion work? I am not sure if Lean is smart enough to infer `d` from a unification problem like `d.propIO = .out`.

Iris/IrisTest/MonPred.lean
```diff
@@ -0,0 +1,472 @@
...
+#guard_msgs (whitespace := lax) in
+example {α : Type _} (Φ : α → PROP) :
+    □ (∀ a, ⎡Φ a⎤) ⊣⊢@{MonPred I PROP} ⎡□ (∀ a, Φ a)⎤ := by
+  istart PROP
+  trace_state
+  iintro %i
+  isplit <;> iintro _ //
+
+-- Expected: `AsEmpValid .from _ .in PROP bi .out iprop(∀ i, 𝓟 -∗ 𝓠)`,
+-- i.e. `asEmpValid_monPred_at_wand` (not `asEmpValid_monPred_at`).
```
MackieLoeffel: What is this comment about?

MackieLoeffel (approved):
Except for some minor comments, this looks good to me. (Thought I am not super familiar with these instances as I have not used monotone predicates myself. I guess we will see if this actually works when someone tries to use monotone predicates with Iris Lean.)

Iris/IrisTest/MonPred.lean
alvinylt: Duplicate comment for the test below. Removed in commit https://github.com/leanprover-community/iris-lean/pull/677/commits/fea2928a2d73028d36bafe6764c56227ecc8c155.

Iris/Iris/ProofMode/InstancesLater.lean
alvinylt: No, they were necessary only when `IntoExcept0` involved semi-output parameters. The config for these three instances and `isExcept0_monPred_at` have been removed.

Commit: https://github.com/leanprover-community/iris-lean/commit/7f93794dc04ddadb51ad955339b249df39c012fb

Iris/Iris/ProofMode/InstancesLater.lean
alvinylt: Oops, I meant "the three non-recursive instances require `set_option synthInstance.checkSynthOrder false` as `◇ P` as an input pins `P` by unification", but somehow it is cut off. Nonetheless, since this is no longer relevant, the comment is removed.

Commit: https://github.com/leanprover-community/iris-lean/commit/7f93794dc04ddadb51ad955339b249df39c012fb

Iris/Iris/ProofMode/InstancesEmbedding.lean
alvinylt: Removed in commit https://github.com/leanprover-community/iris-lean/commit/b9ef2f55882d8419df1d9c871b4355a74287fe1d.

Iris/Iris/ProofMode/MonPred.lean
alvinylt: You're correct about this being problematic.

With this formulation of the instance, the type class synthesiser is unable to apply the instance in the following test.

```lean4
#ipm_synth ∀ 𝓟 (j : I.car), FromAssumption true .in 𝓟 ((iprop(⎡𝓟⎤) : MonPred I PROP).monPred_at j)
```

The solution is to do it the other way round, that is, having `io` for `FromAssumption` which is then mapped to the `MakeMonPredAt.Kind` value for the `MakeMonPredAt` premise.

Commit: https://github.com/leanprover-community/iris-lean/commit/ac0aa4a7436c80cc33c07222630a4ae7fe3de11b

Iris/Iris/ProofMode/MonPred.lean
alvinylt: Without this, a few proofs in the codebase that use `iapply` would be broken.

An example is the following:

```lean4
@[rocq_alias bupd_alt_plainly]
theorem BUpdPlain_plainly {P : PROP} : BUpdPlain iprop(■ P) ⊢ (■ P) := by
  unfold BUpdPlain
  iintro H
  iapply H
  iapply wand_rfl
```

The last line would give the error `iapply: cannot apply MonPred.monPred_at ?m.72 ?m.121 to iprop(■ P -∗ ■ P)`.

The instance `asEmpValid_monPred_at_wand` is tried, which in turn uses `MakeMonPredAt` (`makeMonPredAt_default`), and that is why we get `MonPred.monPred_at ?m.72 ?m.121` I think it is again due to the difference between `Hint Mode` in Rocq and parameter configuration for type classes in Lean. In Rocq, the default instance of `MakeMonPredAt` is never reached, whereas in Lean a value containing metavariables can be returned, which is why we perform the check using the guard.

Iris/Iris/ProofMode/MonPred.lean
```diff
@@ -0,0 +1,1123 @@
...
+        (inst.as_emp_valid_0.as_emp_valid.left hd hφ) i).trans (instMP i).make_monPred_at.mp
+    · refine fun hd hP =>
+        inst.as_emp_valid_0.as_emp_valid.right hd ((monPred_at_emp_valid P).mpr fun i => ?_)
+      calc
+        _ ⊢ ∀ a, Φ a       := hP
+        _ ⊢ Φ i            := forall_elim i
+        _ ⊢ P.monPred_at i := (instMP i).make_monPred_at.mpr
+
+@[ipm_backtrack, rocq_alias as_emp_valid_monPred_at_wand]
+instance asEmpValid_monPred_at_wand (d : AsEmpValid.Direction) (φ : Prop) io
```
MackieLoeffel: Should this instance and the other instances only work when `io = .in`? it seems a bit weird to try this instance if `io = .out` since then somehow the search would need to find the right `PROP`. The same probably also applies to the ProofMode.asEmpValid_embed instance.

MackieLoeffel: I also think these instances should have a priority lower than default, e.g. something like `default - 50`.

Iris/Iris/ProofMode/MonPred.lean
MackieLoeffel: ~~This is weird, I don't understand what is going on here. Should the input checking in the ipm TC synth not just check for mvars at the top level but also if the input is an mvar applied to some arguments? Would this change anything here?~~

Edit: No I don't think it would help here. Maybe just making these instances lower priority than the default is the right solution here? This might also be how it works in Rocq.

alvinylt: There is no specific priority set in Rocq, but a lower priority for `asEmpValid_monPred_at_wand` and `asEmpValid_monPred_at_equiv` indeed avoids the need for the guard. The proofs then prefers `asEmpValid_of_asEmpValid0` and if it fails, it backtracks and tries `asEmpValid_monPred_at_wand_*`.

Commit: https://github.com/leanprover-community/iris-lean/pull/677/commits/1be26a62f37249aaec0a166f50ad006e7c870611

Iris/Iris/ProofMode/MonPred.lean
alvinylt: Same fix as in https://github.com/leanprover-community/iris-lean/pull/677#discussion_r3854979467.

alvinylt: Yes, it works with `io := .in`. Is it accurate to say that `asEmpValid_embed` and `asEmpValid_monPred_at*` are applicable only when `istart PROP` is used (where `PROP` is explicitly specified by the user`), and that is why it must be an input?

Commit: https://github.com/leanprover-community/iris-lean/pull/677/commits/792f304ec440352e7b7b8f9913e44f118a09e607

Iris/Iris/ProofMode/MonPred.lean
MackieLoeffel: Priority in Rocq is weird: It is based on the number of hypotheses of the instance and hypotheses with less arguments are tried first. So the monPred instances implicitly are tried after the asEmpValid0 instance.

Iris/Iris/ProofMode/MonPred.lean
MackieLoeffel: They are also valid for `ihave / iapply`, but there we also have `io := .in`.

PR #680, opened 2026-08-26
feat: add findECtxRev

ayhon:
## Description
Adds a `findECtxRev` function which works like `findECtx`, but explores the expression from the innermost context first. Also adds comments to distinguish `findECtx` and `findECtxRev`.

## Checklist

### DIFF
--- a/Iris/Iris/HeapLang/Tactic.lean
+++ b/Iris/Iris/HeapLang/Tactic.lean
@@ -57,6 +57,12 @@ where
   | e =>
     extractEctxItem e
 
+public meta partial
+def extractAllEctxItems (e : Q(Exp)) (acc : List Q(ECtxItem) := []) : MetaM (List Q(ECtxItem) × Q(Exp)) := do
+  match ← extractEctxItem e with
+  | (.some Ki, e') => extractAllEctxItems e' (Ki :: acc)
+  | (.none, e) => return (acc, e)
+
 open ECtxItem in
 meta partial
 def fillItem (e : Q(Exp)) : Q(ECtxItem) → MetaM Q(Exp)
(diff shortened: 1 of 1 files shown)


Iris/Iris/HeapLang/Tactic.lean
```diff
@@ -112,6 +118,9 @@ structure ECtxResultOf (e : Q(Exp)) (α : Type) where unsafeMk ::
   e' : Q(Exp)
   heq : ProgramLogic.fill $K $e' =Q $e := ⟨⟩
 
+/-- Given an expression `ogE`, finds the *outermost* evaluation context `K` and
+    corresponding expression `e'` such that `K[e'] = e` and `pred K e'` does 
+    not fail -/
```
MackieLoeffel:
```suggestion
    not fail. This corresponds to `reshape_expr` in Rocq. -/
```

Iris/Iris/HeapLang/Tactic.lean
```diff
@@ -123,3 +132,19 @@ where
...
+    (pred : Q(List ECtxItem) → Q(Exp) → ProofModeM α) :
+    ProofModeM (Option (ECtxResultOf ogE α)) := do
+  let (Kis, inner) ← extractAllEctxItems ogE
+  let K : Q(List ECtxItem) := quoteList Kis
+  go inner K
+where
+  go (e' : Q(Exp)) (K : Q(List ECtxItem)) : ProofModeM (Option (ECtxResultOf ogE α)) := do
+    if let some a ← observing? <| pred K e' then
+      return some {result := a, K, e'}
+    let mkApp2 (.const ``List.cons _) Ki K := K | return none
```
MackieLoeffel: Can / should this be an `let_expr`?

MackieLoeffel (approved):
Looks good to me, modulo two small comments.

Iris/Iris/HeapLang/Tactic.lean
ayhon: Yes, I think you're right. This should be a `let_expr` in this case. 

I have grown used to not using `let_expr` since then one can't give Qq type annotations to the pattern match. I'm thinking of something like ```let mkApp2 (.const ``List.cons _) (Ki : Q(ECtxItem)) (K : Q(List ECtxItem)) := ...```. In this case it doesn't matter though, since `Ki` and `K` are just immediately given to the recursive call. So having a `let_expr` just makes it easier to read.

PR #572, opened 2026-08-06
refactor: `Hyps` optimisations (experiment)

alvinylt:
## Description

There are many ways to potentially optimise the Iris Proof Mode, especially regarding how the spatial and intuitionistic contexts are organised using [`Hyps` in `ProofMode/Expr.lean`](https://github.com/leanprover-community/iris-lean/blob/3d3dfe0/Iris/Iris/ProofMode/Expr.lean#L129-L135). To name a few:
- The functions `Hyps.spatialIVarIds` and `Hyps.intuitionisticIVarIds` use list concatenation (`++`): quadratic complexity due to repeated copying of lists. This can easily be optimised using an accumulator (commit [24db6ec](https://github.com/leanprover-community/iris-lean/commit/24db6ecf637c4515bd620caa5963cf654284d4ce)).
- The function [`Hyps.add`](https://github.com/leanprover-community/iris-lean/blob/3d3dfe0/Iris/Iris/ProofMode/Expr.lean#L159-L165) does not perform balancing. It will be interesting to see whether having a balanced tree enables notable optimisations. Otherwise, we might instead simply have two separate lists (as in Rocq), one for spatial hypotheses and another for intuitionistic hypotheses. This would avoid complications (e.g. in [`Hyps.buildAccuProof`](https://github.com/leanprover-community/iris-lean/blob/3d3dfe0/Iris/Iris/ProofMode/Expr.lean#L227-L250)) such as having multiple trees representing the same context (with the same order of hypotheses).
- Possibly some data structures for more efficient lookup of hypotheses (e.g. using [`Hyps.select`](https://github.com/leanprover-community/iris-lean/blob/3d3dfe0/Iris/Iris/ProofMode/Expr.lean#L264-L269)) instead of repeated `isDefEq`.
- Prevent hypotheses from shuffling around when IPM tactics are used.

It will also be helpful to think through how the optimisations can be evaluated.

## Checklist
* [ ] My code follows the mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) and [code style](https://leanprover-community.github.io/contribute/style.html) conventions
* [ ] I have added my name to the `authors` section of any appropriate files

PR #684, opened 2026-08-27
fix: hypothesis order for framing

alvinylt:
## Description

When `iFrame` in Iris-Rocq is used without specifying the hypotheses to be framed, all spatial hypotheses in the context are tried in *reverse* order. It is desirable for `iframe` in Lean to follow the same convention.

### First Solution

Attempted in commit https://github.com/leanprover-community/iris-lean/commit/78724095face181cdf2e5082cbd388da9bdf3e47.

This solution was the simplest. We check whether `iframe` is used with explicit hypotheses. If not, all spatial hypotheses are chosen, and the list is reversed.

There are two problems with this approach.
- `iframe` is no longer the same as `iframe ∗`.
- This addresses the tactic `iframe` but not other ways to apply framing (e.g., the specialisation pattern `[$]`).

### Second Solution

Attempted in commit https://github.com/leanprover-community/iris-lean/commit/66727f5e59f013394788ef69f92c0ba13f0b3d91.

`Hyps.add` is modified to insert a new hypothesis as the *right* child of `.mkSep` instead of the *left* child, resulting in the hypothesis trees being mirrored throughout the codebase. The IPM display (`ProofMode/Display.lean`) is then modified to print the hypotheses in reverse order so that the IPM user sees the hypothesis in the right order.

However, we run into problems with this approach, as other tactics (e.g. `iloeb`, `imod`) start producing hypotheses in different shapes (e.g., `P -∗ Q -∗ R` becomes `Q -∗ P -∗ R`). This breaks a few proofs across the codebase, and fixing them would result in the proofs diverging from their Rocq counterparts. Other problems also emerge as the proposition representing the context becomes left-associative, and that requires reimplementing `iaccu`.

With  [`env_to_list`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/environments.v#L34-35) in Iris-Rocq, the *newest* hypothesis comes first. Meanwhile, [`env_to_prop`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/environments.v#L84-87) and [`env_dom`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/environments.v#L39-40) have the *oldest* hypothesis being the leftmost hypothesis. By flipping `Hyps.add`, the implementation in Lean would move away from what `env_to_prop` is doing. Removing the newest hypothesis (using `Hyps.remove`) would require walking through the entire context, as opposed to the current constant-time operation.

Most crucially, the tactic `iRevert` is [implemented using `iElaborateSelPat`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/ltac_tactics.v#L710), which reverses the pattern order, while [`_iFrame0`](https://gitlab.mpi-sws.org/iris/iris/-/blob/master/iris/proofmode/ltac_tactics.v#L481) parses the selection pattern without reversing the order. So the key is to fix the implementation of `iframe` instead of changing `Hyps`.

### Third (and Current) Solution

Commit https://github.com/leanprover-community/iris-lean/commit/547e83adf8f6d7c136738c084b4ee86c3ba1dbac.

I think this closely resembles what `iFrame` in Rocq is doing.

When selection patterns are parsed (using `SelPat.resolveOne`), we specify whether the wildcard patterns (`∗`, `#`, `%`) should give the hypotheses in reverse order, which is the case only for framing.

This approach also ensures `iframe` and `iframe ∗` are identical.

## Checklist

### DIFF
--- a/Iris/Iris/HeapLang/Lib/Quicksort.lean
+++ b/Iris/Iris/HeapLang/Lib/Quicksort.lean
@@ -201,10 +201,10 @@ theorem quicksort_spec l ls :
     icases Hl with ⟨%l, %tl, %rfl, Hpt, Hl⟩
     wp_load
     wp_smart_apply partition_spec $$ [$] with %l1 %l2 ⟨Hl1, Hl2⟩
-    wp_smart_apply IH $$ [$] with %l1' %ls1' ⟨Hl1, %_, %_⟩
-    wp_smart_apply IH $$ [$] with %l2' %ls2' ⟨Hl2, %_, %_⟩
-    wp_smart_apply cons_spec $$ Hl2 with %_ _
-    wp_smart_apply append_spec $$ [$] with %_ _
+    wp_smart_apply IH $$ [$Hl1] with %l1' %ls1' ⟨Hl1, %_, %_⟩
+    wp_smart_apply IH $$ [$Hl2] with %l2' %ls2' ⟨Hl2, %_, %_⟩
+    wp_smart_apply cons_spec $$ Hl2 with %_ Hcons
+    wp_smart_apply append_spec $$ [$Hl1 $Hcons] with %_ _
     iapply HΦ; iframe; isplit <;> ipureintro
     · have : ls2'.all (head < ·) := by grind
       grind [pairwise_cons]
--- a/Iris/Iris/Instances/Lib/CInvariants.lean
+++ b/Iris/Iris/Instances/Lib/CInvariants.lean
@@ -84,9 +84,8 @@ theorem own_valid {γ : GName} {q1 q2 : Qp} :
     ⊢@{IProp GF} own γ q1 -∗ own γ q2 -∗ ⌜(q1 + q2).val ≤ 1⌝ := by
   unfold own
   iintro H1 H2
-  ihave H := iOwn_op $$ [H1 H2]; iframe H1 H2
-  ihave H := iOwn_cmraValid $$ H
-  icases internalCmraValid_discrete $$ H with %H
+  icombine H1 H2 as H
+  ihave %H := iOwn_cmraValid $$ H
   ipureintro
   exact H.2
 
--- a/Iris/Iris/Instances/Lib/CInvariants.lean
+++ b/Iris/Iris/Instances/Lib/CInvariants.lean
@@ -84,7 +84,7 @@ theorem own_valid {γ : GName} {q1 q2 : Qp} :
     ⊢@{IProp GF} own γ q1 -∗ own γ q2 -∗ ⌜(q1 + q2).val ≤ 1⌝ := by
   unfold own
   iintro H1 H2
-  ihave H := iOwn_op $$ [H1 H2]; iframe
+  ihave H := iOwn_op $$ [H1 H2]; iframe H1 H2
   ihave H := iOwn_cmraValid $$ H
   icases internalCmraValid_discrete $$ H with %H
   ipureintro
--- a/Iris/Iris/ProgramLogic/EctxLifting.lean
+++ b/Iris/Iris/ProgramLogic/EctxLifting.lean
@@ -145,7 +145,6 @@ theorem wp_lift_atomic_base_step_no_fork_fupd (h : toVal e₁ = none) :
   subst h
   simp only [List.length_nil, Nat.add_zero, Algebra.BigOpL.bigOpL_nil]
   iframe
-  itrivial
 
 @[rocq_alias wp_lift_atomic_base_step_no_fork]
 theorem wp_lift_atomic_base_step_no_fork (h : toVal e₁ = none) :
--- a/Iris/Iris/ProofMode/Patterns/SelPattern.lean
+++ b/Iris/Iris/ProofMode/Patterns/SelPattern.lean
@@ -71,7 +71,7 @@ structure SelTarget where
   Resolve selection patterns to concrete proofmode hypotheses (`.ipm`) and pure
   local hypotheses (`.pure`).
 -/
-def SelPat.resolveOne (hyps : Hyps bi e) (reverseWildcardOrder : Bool) :
+def SelPat.resolveOne (hyps : Hyps bi e) (wildcardOrder : HypsOrder) :
     SelPat → ProofModeM (List SelTarget)
   | .ident name => do
       let ivar ← hyps.findWithInfo name
@@ -81,11 +81,11 @@ def SelPat.resolveOne (hyps : Hyps bi e) (reverseWildcardOrder : Bool) :
       addLocalVarInfo name (← getLCtx) ldecl.toExpr ldecl.type
       return [⟨.pure ldecl.fvarId, true⟩]
   | .intuitionistic =>
-      let ivars := hyps.intuitionisticIVarIds
-      return (if reverseWildcardOrder then ivars.reverse else ivars).map (⟨.ipm ·, false⟩)
+      let ivars := hyps.intuitionisticIVarIds wildcardOrder
+      return ivars.map (⟨.ipm ·, false⟩)
   | .spatial =>
-      let ivars := hyps.spatialIVarIds
-      return (if reverseWildcardOrder then ivars.reverse else ivars).map (⟨.ipm ·, false⟩)
+      let ivars := hyps.spatialIVarIds wildcardOrder
+      return ivars.map (⟨.ipm ·, false⟩)
   | .pure => do
       -- `%` selects user-facing Lean pure assumptions, so we keep only `Prop` hypotheses.
       let mut hyps := #[]
@@ -95,7 +95,10 @@ def SelPat.resolveOne (hyps : Hyps bi e) (reverseWildcardOrder : Bool) :
         if ! (← isProp ldecl.type) then
           continue
         hyps := hyps.push (⟨.pure ldecl.fvarId, false⟩)
-      return (if reverseWildcardOrder then hyps.reverse else hyps).toList
+      hyps := match wildcardOrder with
+      | .topToBottom => hyps
+      | .bottomToTop => hyps.reverse
+      return hyps.toList
 
 /--
   Resolve a list of selection targets.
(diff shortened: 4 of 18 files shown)


Iris/IrisTest/Tactics.lean
```diff
@@ -1460,7 +1460,7 @@ example [BI PROP] (P Q R : PROP) : P ⊢ P -∗ R -∗ (P ∗ P -∗ R -∗ Q) -
 
 /-- Tests `ispecialize` with even more complex autoframe. -/
 example [BI PROP] (P : Nat → PROP) (Q R : PROP) :
-    P 1 ⊢ □ P 1 -∗ P 2 -∗ R -∗ (∀ n, ((□ P n ∗ R ∗ P n) -∗ P 2 -∗ Q)) -∗ Q := by
+    P 2 ⊢ □ P 1 -∗ P 1 -∗ R -∗ (∀ n, ((□ P n ∗ R ∗ P n) -∗ P 2 -∗ Q)) -∗ Q := by
   iintro HP1 #HP1' HP2 HR HPQ
   ispecialize HPQ $$ [$] [$]
```
alvinylt, line 1465: With `HPQ : ∀ n, □ P n ∗ R ∗ P n -∗ P 2 -∗ Q`, the first `[$]` refers to auto-framing for `□ P ?n ∗ R ∗ P ?n`, where the metavariable `n` is instantiated as an effect of the hypothesis being used for framing.

Originally, the older hypothesis (`HP1`) is used, so `HPQ` becomes `□ P 1 ∗ R ∗ P 1 -∗ P 2 -∗ Q`, and the premises are discharged. With the framing order reversed, `HPQ` becomes `□ P 2 ∗ R ∗ P 2 -∗ P 1 -∗ Q` instead.

Hence, the change is necessary for the test to make sense.

Iris/Iris/ProofMode/Tactics/Specialize.lean
```diff
@@ -170,7 +170,9 @@ private def finishSubgoal {u} {prop : Q(Type u)} {bi : Q(BI $prop)} {e}
     return ⟨el, hypsl, q($(pf').mp.trans <| sep_mono_right $pf'')⟩
   -- Auto-framing: `[$]`, `[#$]` and `[>$]`
   | none =>
-    let res ← iFrame hyps goal <| ← SelPat.resolve hyps [.spatial, .intuitionistic]
+    let res ←
+      (SelPat.resolve hyps [.spatial, .intuitionistic] (reverseWildcardOrder := true)) >>=
+      (iFrame hyps goal ·)
```
alvinylt, line 175: For the specialisation pattern `[$]`.

Iris/Iris/ProofMode/Tactics/Intro.lean
```diff
@@ -188,7 +188,7 @@ partial def iIntroCore {u} {prop : Q(Type u)} {bi : Q(BI $prop)}
           fun hyps' goal' fvars => withoutFVars (u := 0) fvars
             <| iIntroCore hyps' goal' ((ref, .clear selPats) :: pats) k
       | ⟨true, s⟩ :: selPats =>
-        let res ← s.resolveOne hyps >>= iFrame hyps Q
+        let res ← s.resolveOne hyps (reverseWildcardOrder := true) >>= iFrame hyps Q
```
alvinylt: For the introduction pattern `{$…}` that clears hypotheses by framing.

Iris/Iris/ProofMode/Tactics/Frame.lean
```diff
@@ -157,7 +157,7 @@ elab "iframe " pats:(colGt ppSpace selPat)+ : tactic => do
   let pats ← liftMacroM <| SelPat.parse pats
 
   ProofModeM.runTactic `iframe λ mvar { hyps, goal, .. } => do
-    let pats ← SelPat.resolve hyps pats
+    let pats ← SelPat.resolve hyps pats (reverseWildcardOrder := true)
```
alvinylt: For `iframe`.

Iris/Iris/ProgramLogic/EctxLifting.lean
```diff
@@ -145,7 +145,6 @@ theorem wp_lift_atomic_base_step_no_fork_fupd (h : toVal e₁ = none) :
   subst h
   simp only [List.length_nil, Nat.add_zero, Algebra.BigOpL.bigOpL_nil]
   iframe
-  itrivial
```
alvinylt, line 148: This one is interesting. The proof state before calling `iframe` is:
```lean4
∗x✝ : stateInterp x✝ (ns + 1) obs' nt
∗x✝¹ : ∃ v, ⌜toVal x✝¹ = some v⌝ ∧ Φ v
⊢ |={E₁}=> stateInterp x✝ (ns + 1) obs' nt ∗ (∃ v, ⌜toVal x✝¹ = some v⌝ ∧ Φ v) ∗ emp
```

When the hypothesis was handled in forward order, `stateInterp x✝ (ns + 1) obs' nt` is first framed. So the proof state became:

```lean4
∗x✝¹ : ∃ v, ⌜toVal x✝¹ = some v⌝ ∧ Φ v
⊢ |={E₁}=> (∃ v, ⌜toVal x✝¹ = some v⌝ ∧ Φ v) ∗ emp
```

Then, the second framing step gives the proof state:

```lean4
⊢ |={E₁}=> emp
```

That's why `itrivial` is needed.

Now with the framing in reverse order, the `emp` is removed after the first framing step, so the goal is closed directly.

Iris/Iris/Instances/Lib/CInvariants.lean
```diff
@@ -84,7 +84,7 @@ theorem own_valid {γ : GName} {q1 q2 : Qp} :
     ⊢@{IProp GF} own γ q1 -∗ own γ q2 -∗ ⌜(q1 + q2).val ≤ 1⌝ := by
   unfold own
   iintro H1 H2
-  ihave H := iOwn_op $$ [H1 H2]; iframe
+  ihave H := iOwn_op $$ [H1 H2]; iframe H1 H2
```
alvinylt: Some arguments of `iOwn_op` are not explicitly specified and thus metavariables are left for `iframe` to unify.

```lean4
GF : BundledGFunctors
W : CInvG GF
γ : GName
q1 q2 : Qp
⊢ 
∗H1 : iOwn γ (none, some (DFrac.own q1))
∗H2 : iOwn γ (none, some (DFrac.own q2))
⊢ iOwn ?m.66 ?m.67 ∗ iOwn ?m.66 ?m.68
```

Iris/Iris/HeapLang/Lib/Quicksort.lean
```diff
@@ -201,10 +201,10 @@ theorem quicksort_spec l ls :
     icases Hl with ⟨%l, %tl, %rfl, Hpt, Hl⟩
     wp_load
     wp_smart_apply partition_spec $$ [$] with %l1 %l2 ⟨Hl1, Hl2⟩
-    wp_smart_apply IH $$ [$] with %l1' %ls1' ⟨Hl1, %_, %_⟩
-    wp_smart_apply IH $$ [$] with %l2' %ls2' ⟨Hl2, %_, %_⟩
-    wp_smart_apply cons_spec $$ Hl2 with %_ _
-    wp_smart_apply append_spec $$ [$] with %_ _
+    wp_smart_apply IH $$ [$Hl1] with %l1' %ls1' ⟨Hl1, %_, %_⟩
```
alvinylt, line 204: The induction hypothesis is:

```lean4
□IH :
∀ l ls Φ,
  isList l ls -∗
    (▷ ∀ l' ls', isList l' ls' ∗ ⌜Pairwise LE.le ls'⌝ ∗ ⌜ls ~ ls'⌝ -∗ Φ l') -∗ WP hl(v(&quicksort) v(&l)) {{ Φ }}
```

When the universally quantified `l` is instantiated with `l2`, the subsequent steps fail. Hence we stipulate the use of `Hl1` to unify `l` with `l1`.

alvinylt (commented):
Here are some remarks regarding the modifications.

Iris/Iris/ProofMode/Patterns/SelPattern.lean
```diff
@@ -71,7 +71,8 @@ structure SelTarget where
   Resolve selection patterns to concrete proofmode hypotheses (`.ipm`) and pure
   local hypotheses (`.pure`).
 -/
-def SelPat.resolveOne (hyps : Hyps bi e) : SelPat → ProofModeM (List SelTarget)
+def SelPat.resolveOne (hyps : Hyps bi e) (reverseWildcardOrder : Bool) :
```
MackieLoeffel: Instead of making `reverseWildcardOrder` an optional boolean, I think it is better to make it a mandatory parameter with a type like the following:
```
Inductive WildcardOrder
  | topToBottom | bottomToTop
```
I think this parameter should be passed to `intuitionisticIVarIds` and `spatialIVarIds`, which should do the reversing if necessary (or construct the list in the other way around in the first place). This makes the order more explicit.

MackieLoeffel (commented):
Thanks for looking into this! I think your proposed solution makes sense, but I think it would be better to be more explicit about the wildcard direction. See my comment.

Iris/Iris/ProofMode/Patterns/SelPattern.lean
alvinylt: The inductive structure is introduced into `ProofMode/Expr.lean` (named `HypsOrder` instead of `WildcardOrder` as `Hyps.spatialIVarIds` and `Hyps.intuitionisticIVarIds` are generic functions, not just for selection patterns).

The parameter is no longer optional, so all call sites of `SelPat.resolve`/`SelPat.resolveOne` have to be updated.

Commit: https://github.com/leanprover-community/iris-lean/pull/684/commits/5d35422ba2255c2714aeaaf612424392348e15bc

alvinylt: Besides this, there is an easy optimisation for these two functions (#572), which might as well be adopted here.

Commit: https://github.com/leanprover-community/iris-lean/pull/684/commits/6837509f8d2f6cf6b187fae2e36e20b7f7fe4a0b

Iris/Iris/ProofMode/Tactics/Frame.lean
```diff
@@ -157,7 +157,7 @@ elab "iframe " pats:(colGt ppSpace selPat)+ : tactic => do
   let pats ← liftMacroM <| SelPat.parse pats
 
   ProofModeM.runTactic `iframe λ mvar { hyps, goal, .. } => do
-    let pats ← SelPat.resolve hyps pats
+    let pats ← SelPat.resolve hyps pats .bottomToTop
```
MackieLoeffel, line 161:
```suggestion
    -- .bottomToTop since we want to frame the most recently introduced hypotheses first, matching Iris-Rocq
    let pats ← SelPat.resolve hyps pats .bottomToTop
```

MackieLoeffel (approved):
Looks great! I am curious whether this has an impact on the bench, but otherwise this is ready to be merged.

PR #686, opened 2026-08-29
chore: Standardize the per-file copyright notices.

markusdemedeiros:
This is a minor thing that has bothered me for a while. This PR changes all of the headers to have a uniform copyright header. I preserved the `Authors` field, but it has no impact on copyright. 

As far as I can tell this is within the parameters of the Apache 2.0 license, as long as we make a git tag with the old headers before merging this (see the new file `NOTICE`). The license says they need to be preserved, but preserving a snapshot of them in the history should be fine.

### DIFF
--- a/Iris/Iris.lean
+++ b/Iris/Iris.lean
@@ -1,7 +1,6 @@
 /-
 Copyright (c) The Iris-Lean Contributors
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors:
 -/
 module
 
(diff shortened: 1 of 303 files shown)


MackieLoeffel (approved):
Good idea to clean this up. My understanding is also that this is fine copyright-wise. I am not fully sure if the notice file is even necessary, but I guess it does not hurt to have it.

Kaptch (approved):

lzy0505 (approved):

